PR #4, opened 2020-12-16
Fix default setup code in CLI docs

aaaidan:
This fixes what looks like a small typo in the CLI docs.

I updated to match code in [`Settings.h`](https://github.com/HomeSpan/HomeSpan/blob/8b4d4cf32dc33afdc80b4268abc2559d08afcd8d/src/Settings.h#L46)

(Full disclosure: I haven't tested this default code works.)

Aside from this, it might make the developer "out of box" easier if the default code was also listed on a more prominent page. I notice it's in the end user guide, but developers may not read that before trying to get up and running. (I didn't!)


HomeSpan:
Thanks for finding this - it is indeed a typo.  I'll also find a place to incorporate this into the Getting Started page so it is more prominent per your suggestion.   Though the default code looks arbitrary, it's actually the numbers represented if you try to spell the word "homespan" on a telephone keypad.

aaaidan:
Very nice. I assumed it was something memorable to someone, but that's fun.

PR #28, opened 2021-02-03
Added a small section to docs

### LINKED ISSUE #27: Make it possible to preset client SSID and password
As far as I can see it is not possible to set the SSID and password (for the wifi where the device will finally work in) via code. Like the `homeSpan.setApSSID(ssid)` method I propose `homeSpan.setSSID(ssid)` as well as for the passwords.

flashspys:
Explained why its not possible to define passwords in code. Closes #27


HomeSpan:
I like the idea of making the WiFi question more explicit but am not sure the User Guide is the correct place to add this since the focus of the User Guide is on using HomeSpan without the Arudino IDE.  Instead, I created a new [FAQ](https://github.com/HomeSpan/HomeSpan/blob/master/docs/FAQ.md) page and added this as the first question because I agree its important and hopefully will prevent future confusion.  Thanks for prompting this!

PR #59, opened 2021-03-07
Add TargetHeaterCoolerState Properly

jakeholland:
`TargetHeaterCoolerState` was not present in the service list.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .development
+.DS_Store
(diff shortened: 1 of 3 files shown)


HomeSpan:
Thanks for catching this.  By coincidence I've just finished re-working this section of the code to include more error-checking for out-of-range issues that could (almost always) cause a device to become non-responsive.   While re-working the code I added back the TargetHeaterCoolerState.  You can see the latest changes in the dev-1.2.1 branch, which I hope to release this weekend.

As an aside, I originally included TargetHearerCoolerState but had to remove it because the HAP document did not include the parameters (such as the UUID).  However, I later realized the HAP document has some formatting issues and this Characteristic, which found on page 223 of HAP, has its data on page 228!

HomeSpan:
This has been fixed in release 1.2.1.

PR #121, opened 2021-08-03
add support for Televisions

unreality:
This pull adds support for the following services:
* Televisions
* TelevisionSpeaker
* InputSource

And the following Characteristics:
* ActiveIdentifier
* ClosedCaptions
* ConfiguredName
* CurrentMediaState
* Identifier
* InputDeviceType
* InputSourceType
* PictureMode
* PowerModeSelection
* RemoteKey
* SleepDiscoveryMode
* TargetMediaState
* TargetVisibilityState
* VolumeControlType
* VolumeSelector

Also adds an example for a sample Television

### DIFF
--- a/examples/19-Television/19-Television.ino
+++ b/examples/19-Television/19-Television.ino
@@ -1,0 +1,125 @@
+#include "HomeSpan.h"
+
+const int OTHER = 0;
+const int HOME_SCREEN = 1;
+const int TUNER = 2;
+const int HDMI = 3;
+const int COMPOSITE_VIDEO = 4;
+const int S_VIDEO = 5;
+const int COMPONENT_VIDEO = 6;
+const int DVI = 7;
+const int AIRPLAY = 8;
+const int USB = 9;
+const int APPLICATION = 10;
+
+struct HKTV : Service::Television {
+
+  SpanCharacteristic *active;
+  SpanCharacteristic *activeIdentifier;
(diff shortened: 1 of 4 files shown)


HomeSpan:
Sorry for the delay in responding to this, and thanks very much for putting this together.  Is there a (public) document available with the actual HAP specs for TV's?

unreality:
No public document that i'm aware of, I think the only official one is the commercial HAP document that is under NDA. The IDs, Characteristic names, etc I found looking at the HAP-NodeJS repository.

unreality:
Any progress on getting this merged? Im happy to keep my fork with the TV additions but it would be nicer for them to be merged into the main project i think...

HomeSpan:
I'm currently working through compatibility issues and fixes so HomeSpan will compile and run under Arduino-ESP32 2.0.0 (released a few weeks ago) with support for the ESP32-S2 and ESP32-C3 chips (which are slowly replacing the ESP32 chip).  Small differences between the chips are just enough to "break" certain HomeSpan feature. My plan is to weave the TV controls into a new 2.0 version of HomeSpan that supports all three ESP32 chips.

unreality:
Essentially, yes.

* I also couldn't get the TV Volume to work correctly, but its possible i'm doing it the wrong way.
* PowerModeSelector I believe refers to 'Power Saving' type features

PR #327, opened 2022-06-17
Added isConnected property

batonPiotr:


HomeSpan:
Please use HomeSpan's WiFi callback feature to determine when WiFi has been connected.  HomeSpan internally handles reconnects upon disconnect.

PR #328, opened 2022-06-17
Added public aid getter

batonPiotr:
I'd like to be able to retrieve the `aid` of an accessory in order to be able to remove it from the database later. There is though a possibility to set it on the first place, but I don't want to be forced to manage those IDs internally if there is already a system for this in the library itself.

### DIFF
--- a/src/HomeSpan.h
+++ b/src/HomeSpan.h
@@ -323,6 +323,9 @@ class SpanAccessory{
 
   public:
   
+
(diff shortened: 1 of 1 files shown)


HomeSpan:
Management of AIDs is required if you are going to delete them later since HomeKit requires AIDs to remain consistent as new accessories are added/deleted.

PR #374, opened 2022-09-04
Add PlatformIO compatibility

oleksiikutuzov:
A minor adjustment that makes the library compatible with PlatformIO. Tested on both PlatformIO and Arduino, it seems to be working fine.


HomeSpan:
What does HomeSpan show as the Arduino-ESP32 version number in the Serial Monitor when run under PlatformIO?  I'm not sure it will show the correct value.

oleksiikutuzov:
Yeah, it's showing `Arduino-ESP Ver.: ARDUINO_ESP32_GIT_DESC`

HomeSpan:
I don't have PlatformIO installed but my understanding from others is a problem with the GitHub deployment - it does not include core_version.h.  However, Arduino.h does contain alternate version information, even though it is not always updated correctly.    I can add in some logic to use this information anyway when compiling with PlatformIO to avoid any compile errors.

oleksiikutuzov:
However, it includes esp_arduino_version.h and we can get the version number from there.
https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/esp_arduino_version.h

It worked out, now it shows correctly `Arduino-ESP Ver.: 2.0.3`

HomeSpan:
Yes, this should work, but only for versions 2.0.4 and 2.0.3.  For some reason Espressif forgot to update the version numbers in 2.0.2 and 2.0.1 - they both show 2.0.0 instead.  If not for that, I would not need to use core_version at all since arduino_version.h can be used with PlatformIO as well Arduino.  I'll add this into the next release.

oleksiikutuzov:
I am not sure what has changed, but PlatformIO seems to compile with no errors and pick up the correct Arduino platform version now without these fixes applied. So I think this can be closed as irrelevant for now.

However, I am getting a bunch of errors compiling sketch with Arduino platform version 2.0.2. Seems like now HomeSpan requires at least version 2.0.3. 
```
In file included from .pio/libdeps/esp32/HomeSpan/src/HomeSpan.h:44,
                 from include/DEV_Sensor.hpp:1,
                 from src/main.cpp:52:
.pio/libdeps/esp32/HomeSpan/src/Utils.h:84:10: error: 'touch_value_t' does not name a type; did you mean 'touch_pad_t'?
   static touch_value_t threshold;
          ^~~~~~~~~~~~~
          touch_pad_t
.pio/libdeps/esp32/HomeSpan/src/Utils.h:174:33: error: 'touch_value_t' has not been declared
   static void setTouchThreshold(touch_value_t thresh){threshold=thresh;}
                                 ^~~~~~~~~~~~~
.pio/libdeps/esp32/HomeSpan/src/Utils.h: In static member function 'static boolean PushButton::TRIGGER_ON_TOUCH(int)':
.pio/libdeps/esp32/HomeSpan/src/Utils.h:107:66: error: 'threshold' was not declared in this scope
   static boolean TRIGGER_ON_TOUCH(int pin){return(touchRead(pin)<threshold);}
                                                                  ^~~~~~~~~
.pio/libdeps/esp32/HomeSpan/src/Utils.h: In static member function 'static void PushButton::setTouchThreshold(int)':
.pio/libdeps/esp32/HomeSpan/src/Utils.h:174:55: error: 'threshold' was not declared in this scope
   static void setTouchThreshold(touch_value_t thresh){threshold=thresh;}
                                                       ^~~~~~~~~
.pio/libdeps/esp32/HomeSpan/src/Utils.h:174:55: note: suggested alternative: 'thresh'
   static void setTouchThreshold(touch_value_t thresh){threshold=thresh;}
                                                       ^~~~~~~~~
                                                       thresh
In file included from include/DEV_Sensor.hpp:1,
                 from src/main.cpp:52:
.pio/libdeps/esp32/HomeSpan/src/HomeSpan.h: At global scope:
.pio/libdeps/esp32/HomeSpan/src/HomeSpan.h:720:33: error: 'touch_value_t' has not been declared
   static void setTouchThreshold(touch_value_t thresh){PushButton::setTouchThreshold(thresh);}
                                 ^~~~~~~~~~~~~
```

HomeSpan:
Thanks for catching the issue with touch_value_t.  Looks like this was introduced in 2.0.3.  I incorrectly assumed it was part of the full 2.0 branch.  Will address in next release.

PR #528, opened 2023-03-13
Ability to add 433MHz RF Span Switch

MnemonicFish:
Currently I'm using 2x ESP32's, 2x 16Ch NPN optocouplers (24V to 3.3v) and 2x RSJ485 to encode 32 lights via DMX512 using switches and Homekit App and works great, each light can be manually controlled via the optocoupler wired retractive light switches, however I'm looking to reduce the need for wiring and reduce the number of components. 

I came across a great tutorial on RNT: https://randomnerdtutorials.com/alexa-echo-with-esp32-and-esp8266/

which uses 433MHz RF light switches, RF receiver/ transmitters with the library provided by ([rc-switch](https://github.com/sui77/rc-switch)/RCSwitch.h).  

The business end of the code uses an interrupt. can you advise if there is a way to add this functionality without any adverse affects? 

Code snippets

```
RCSwitch mySwitch = RCSwitch();
```
```
mySwitch.enableReceive(RF_RECEIVER);  // Receiver on interrupt 0 => that is pin #2`
```

```
  if (mySwitch.available()) {    
  /*Serial.print("Received ");
    Serial.print( mySwitch.getReceivedValue() );
    Serial.print(" / ");
    Serial.print( mySwitch.getReceivedBitlength() );
    Serial.print("bit ");
    Serial.print("Protocol: ");
    Serial.println( mySwitch.getReceivedProtocol() );*/

    if (mySwitch.getReceivedValue()==6819768) {
      digitalWrite(RELAY_PIN_1, !digitalRead(RELAY_PIN_1));
    }
    if (mySwitch.getReceivedValue()==9463928) {
      digitalWrite(RELAY_PIN_2, !digitalRead(RELAY_PIN_2));     
    }
    delay(600);
    mySwitch.resetAvailable();
  }
```

is adding RF Span switch to Homespan something that we can expect in the future? 

many thanks for your hard work! 

Fish


HomeSpan:
Glad you are finding HomeSpan useful in controlling lights via an optocoupler.  I'm not sure if the RF Switch library would interfere with any of the existing HomeSpan functionality.   Using interrupts is not necessarily a problem - it depends on exactly what happens during the interrupt, and whether there are delays that cause problems.  The best way to know for sure is to experiment.  If you do get the library working, but it seems to cause delays with HomeSpan responses, you can try using HomeSpan's `autoPoll()` feature to push polling into a separate task (which uses a separate core if running on a dual-core device).  

More generally, it would be possible to extend the functionality of HomeSpan's RF Remote class to also receive, instead of just transmit, RF signals.  I'd had not found a good use case for this, but the RF Switches you note are worth exploring at some point.

Question:  Your comment above is part of a PR, but you did not change any code --- the commits above are part of the dev branch that I'm working on for the next update  Did you mean this to be a PR?

HomeSpan:
Thanks - I'll close this out then.

PR #618, opened 2023-07-27
Fix docu for missing TemperaturSensor

mgeramb:
There was a line break missing which causes that the TemperaturSensor are not shown on the docu page.


HomeSpan:
Thanks - I'll make the change in the next release.

HomeSpan:
Fixed in upcoming 1.8.1 release.

PR #619, opened 2023-07-27
Allow overload limit of Accessories

mgeramb:
In my case a higher limit of Accessories is possible. This PR allows to overload the limit.

### DIFF
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -270,6 +270,7 @@ Creating an instance of this **class** adds a new HAP Accessory to the HomeSpan 
 
   * every HomeSpan sketch requires at least one Accessory
   * a sketch can contain a maximum of 41 Accessories per sketch (if exceeded, a runtime error will the thrown and the sketch will halt)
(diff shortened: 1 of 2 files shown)


HomeSpan:
It's a good idea to allow for more than 41 Accessories since depending on the specific sketch, it may be possible to use more.  However, I'm not sure your fix works.  I assume you intended the user to define HOMESPAN_MAX_ACCESSORIES in the main sketch, but this does not propagate to the library - the user would need to define this in the library code itself, which is not ideal.

I'll add a new homeSpan option to raise this limit (or perhaps try to set the limit automatically based on available memory) and add to the next release.

mgeramb:
I‘am using the Platformio with the arduino framework. This includes the lib as source. Setting the variable before include homespan.h works fine. But an option would be the better solution . Thanks for your help!

HomeSpan:
Alternative solution:  the Accessory Limit of 41 has been changed to 150, which is the maximum allowed by HAP.  In its place HomeSpan now checks the memory low-water mark after initialization, and if it is below DEFAULT_LOW_MEM_THRESHOLD (currently set as 80,000 bytes), a WARNING message is shown on the Serial Monitor.   This limit was determined based on general experimentation but may prove to be too low depending on the specifics of any given sketch.

Code for this is in the "release-1.8.1-dev" branch.

PR #631, opened 2023-08-20
Support the Non-English WiFi name

cccat6:
Specify the charset as UTF-8 so that it will display the Non-English WiFi name in the configuration process correctly.
I tried in my Chinese WiFi name, and it works well.


HomeSpan:
Thanks - will add that to the next release.

clasik (approved):

HomeSpan:
Added to upcoming 1.8.1 release

PR #647, opened 2023-09-16
Add PlatformIO compatibility

andiradulescu:
As discused in https://github.com/HomeSpan/HomeSpan/pull/374 in comment https://github.com/HomeSpan/HomeSpan/pull/374#issuecomment-1236348925 this replaces:
`#include <core_version.h>`
with 
`#include <esp_arduino_version.h>`

### DIFF
--- a/src/Settings.h
+++ b/src/Settings.h
@@ -27,7 +27,7 @@
  
 // USER-DEFINED SETTINGS AND REFERENCE ENUMERATION CLASSES
 
-#include <core_version.h>
+#include <esp_arduino_version.h>
 
 #pragma once
 


HomeSpan:
This change by itself will not work.   HomeSpan.cpp needs to be updated as well to pull version numbers from esp_arduino_version.h instead of core_version.h.   If you run HomeSpan with this PR you will see that the version number is listed as ARDUINO_ESP32_GIT_DESC (the name of the macro that is in core_version.h, but is no longer defined without core_version.h included).

Also, note that Arduino.h already includes esp_arduino_version.h, so no need to include it as a separate line item.

andiradulescu:
I’m working on a fork of a project that uses HomeSpan and I had some issues since I’m using PlatformIO.
Yesterday it wouldn’t work without this change.
Today it works fine without this change.
I’m closing it for now.

PR #657, opened 2023-10-01
Log power and RAM

frankonski:
Logging power level, RAM and PSRAM.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "C_Cpp.default.compilerPath": ""
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "C_Cpp.default.compilerPath": ""
+}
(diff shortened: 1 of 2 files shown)


HomeSpan:
I've given more thought to this and realized there is potentially an infinite number of rows that one may wish to add to the initial table of the Web Log output, which can quickly become overwhelming.  So instead of continuing to add new rows, I created a general callback method so that users can easily add as many new rows with whatever data they'd like in the initial table of the Web Log output.  The way I coded the callback also has the optional benefit of allowing the user to end the initial table and create a new block of custom HTML text that is inserted in the Web Log before the second table is produced.

This code has been added to the 1.8.1 branch (which is serving as the dev branch for the next release).  See the API Reference and Message Logging pages in the 1.8.1  branch for details and examples (including your DRAM example).

On a related note, I'm not sure your PSRAM code would have produced the results you expected.  You would need to subtract the size of the non-PSRAM memory, otherwise it simply provides the total memory even for systems with no PSRAM (there's actually a lot of subtlety in the various ESP32 memory calculations, such as whether you include only DRAM, IRAM, or both DRAM and IRAM, etc.).

In any case, please feel free to test out the new callback functionality in 1.8.1 and let me know what you think.  Hopefully it will be a good generic solution to adding new Web Log data fields.

frankonski:
Yes, I kinda see that coming as well, as everyone could end up wanting to add their own "status data". The doc was not quite accurate to how the API was actually working, so I submitted PR #671.

The callback fits the needs of this PR, so closing.

PR #659, opened 2023-10-02
Preventing possibility of buffer overwrite.

frankonski:
If `u.STRING` is larger than 63 chars, this will cause a buffer overwrite issue.
Dunno if you want to treat this as a CVE or not, as I don't know if you check the string's length somewhere else.

### DIFF
--- a/src/HomeSpan.h
+++ b/src/HomeSpan.h
@@ -490,7 +490,7 @@ class SpanCharacteristic{
         return(String(c));        
       case FORMAT::STRING:
       case FORMAT::DATA:
-        sprintf(c,"\"%s\"",u.STRING);
+        sprintf(c,"\"%.61s\"",u.STRING); // Truncating string to 61 chars.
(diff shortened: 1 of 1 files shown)


HomeSpan:
Actually, I think you identified a slightly deeper bug - I should have been checking for string length further upstream and provided a way to change maxLen if needed.  Also, I added DATA as a new type based on STRING, but the length of DATA can be a lot larger than a STRING so I'll need to provide a method to differentiate the two.

I'll keep this PR open as a placeholder but will close after I address the bug in more detail.

HomeSpan:
After reviewing all the HAP specs about maximum lengths for strings, I think this change to limit the size of the string as you indicated would be the best solution.

HomeSpan:
Will add in PR, though we I will expand buffer to handle a net total of 64 characters, since HAP allows for that without needing to add a new tag.

PR #668, opened 2023-10-21
vTaskDelay in pollTask configurable

mgeramb:
I need fully control of the poll timing in my application. So it would be nice to have the vTaskDelay configurable. Btw, I'am not sure if the poll function should use a delay other than 0. So maybe the poll function should not have the delay parameter and should call pollTask with 0.

### DIFF
--- a/src/HomeSpan.cpp
+++ b/src/HomeSpan.cpp
@@ -153,7 +153,7 @@ void Span::begin(Category catID, const char *displayName, const char *hostNameBa
 
 ///////////////////////////////
 
-void Span::poll() {
+void Span::poll(int taskDelay) {
 
   if(pollTaskHandle){
     LOG0("\n** FATAL ERROR: Do not call homeSpan.poll() directly if homeSpan.start() is used!\n** PROGRAM HALTED **\n\n");
@@ -161,12 +161,12 @@ void Span::poll() {
     while(1);    
   }
   
-  pollTask();
+  pollTask(taskDelay);
 }
 
 ///////////////////////////////
 
-void Span::pollTask() {
+void Span::pollTask(int taskDelay) {
 
   if(!strlen(category)){
     LOG0("\n** FATAL ERROR: Cannot start homeSpan polling without an initial call to homeSpan.begin()!\n** PROGRAM HALTED **\n\n");
@@ -300,8 +300,8 @@ void Span::pollTask() {
   }
 
   statusLED->check();
-
-  vTaskDelay(5);
+  if (taskDelay > 0)
+    vTaskDelay(taskDelay);
     
 } // poll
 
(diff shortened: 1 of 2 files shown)


HomeSpan:
Can you provide some more details about what you are trying to do.  The polling loop takes a variable amount of time depending on a wide variety of variables.  Anything that assumes a specific time may run into trouble later if I add new features in the polling functions or if HomeKit extends functionality, etc.

If you have code that is sensitive to timing, I recommend using autoPoll() to place HomeSpan polling in a separate task, and then you can add your time-sensitive code to the main sketch `loop()` where it will not be blocked by polling, and polling will not be blocked by any delays you may need to add within the `loop()`.

mgeramb:
I use the lib in the [KNX SmartHomeBridge](https://github.com/OpenKNX/OAM-SmartHomeBridge) project. It is combined with a KNX-stack and the [alexa](https://github.com/Aircoookie/Espalexa). Unfortunately, the homespan did not work with the autoPoll function, regardless of the used core. So I decide to control the loop calls by myself and found that the call to loop takes to long, so I found the vTaskDelay which makes now sense if the loop called manually.

HomeSpan:
After re-reviewing this section of the code I realized that I only added the call to `vTaskDelay()` because it is needed when spawning `pollTask()` into a separate task using `homeSpan.autoPoll()` (the task needs to periodically yield to allow other tasks to run).  However, it is NOT needed at all if you are calling `homeSpan.poll()` directly from within the main `loop()`.  So rather than provide a mechanism to change the delay duration, I simply made the use of `vTaskDelay()` conditional on the use of `autoPoll()`. In other words, if you call `poll()` directly, there is no longer any call to `vTaskDelay()`.  If users do want to create a delay in the main `loop()` they can always add `delay()` if needed, though of course this will delay `poll()` and should be use judiciously.  

The change described above is available in the latest development branch: release-1.8.1-dev.  Please give it a try and let me know if it works.

mgeramb:
Tested now and works fine for me. But I'am not sure if your implementation is the best for the readability. Maybe it would be better to add the delay function to the autoPoll function instead of checking the for the taskhandle in the poll function:
```
void autoPoll(uint32_t stackSize=8192, uint32_t priority=1, uint32_t cpu=0){     // start pollTask()
    xTaskCreateUniversal([](void *parms){
      for(;;) 
      { 
        homeSpan.pollTask(); 
        vTaskDelay(5); 
      }}, "pollTask", stackSize, NULL, priority, &pollTaskHandle, cpu);
    LOG0("\n*** AutoPolling Task started with priority=%d\n\n",uxTaskPriorityGet(pollTaskHandle)); 
  }
```

HomeSpan:
That's a good idea and it is more consistent with the way the Arduino-ESP32 library handles the main loop.  Will make the change.

HomeSpan:
Done.  Code is updated in the release-1.8.1-dev branch.  Thanks for the suggestion.

HomeSpan:
Anything more to test on this, or can I close out the PR?

mgeramb:
No, thank you for your help

PR #670, opened 2023-10-27
Verbose WiFi Reconnect

frankonski:
Add the ability to turn off WiFi reconnect messages which may end up clogging WebLogs when WiFi disconnects are expected to happen.


HomeSpan:
Thanks - I've added to release-1.8.1-dev branch.

PR #671, opened 2023-10-29
Correction to doc about new setWebLogCallback()

frankonski:
In response to PR #657, had to adjust the documentation to get the callback working just fine.

### DIFF
--- a/docs/Logging.md
+++ b/docs/Logging.md
@@ -86,23 +86,23 @@ Note that HomeSpan outputs the full content of the Web Log HTML, including whate
 
 Homespan provides a hook into the text used to generate the Web Log that you can extend to add your own data to the initial table as well as more generally add any custom HTML.
 
-To access this text, set a Web Log callback using `homeSpan.setWebLogCallback(void (*func)(String &htmlText))` where
+To access this text, set a Web Log callback using `homeSpan.setWebLogCallback(String (*func)())` where
 
-  * *func* is a function of type *void* that takes a single argument of type *String*, and
-  * *htmlText* will be set by HomeSpan to a String reference containing all the HTML text that the Web Log has already generated to produce the initial table.
+  * *func* is a function returning a *String* that takes no argument.
+  * The string returned will be appended by HomeSpan to the bottom of the initial table.
 
-To add your own data to the table, simply extend the String *htmlText* by adding as many `<tr>` and `<td>` HTML tags as needed.  If you wish to end the table and add any other HTML, simple include the `</table>` tag in *htmlText*, and then add any other custom HTML.  For example, the following function could be used to extend the initial Web Log table to show free DRAM, end the table, and provide a hot link to the HomeSpan Repo:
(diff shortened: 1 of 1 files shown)


docs/Logging.md
```diff
@@ -86,23 +86,23 @@ Note that HomeSpan outputs the full content of the Web Log HTML, including whate
...
 
-To add your own data to the table, simply extend the String *htmlText* by adding as many `<tr>` and `<td>` HTML tags as needed.  If you wish to end the table and add any other HTML, simple include the `</table>` tag in *htmlText*, and then add any other custom HTML.  For example, the following function could be used to extend the initial Web Log table to show free DRAM, end the table, and provide a hot link to the HomeSpan Repo:
+To add your own data to the table, simply have your callback function return a String variable by adding as many `<tr>` and `<td>` HTML tags as needed. Ensure to include the `</table>` tag in your callback to terminate the table. You can then add any other custom HTML.  For example, the following function could be used to extend the initial Web Log table to show free DRAM, end the table, and provide a hot link to the HomeSpan Repo:
 
 ```C++
-void extraData(String &r){
+String extraWebData(){
+  String r = "";
   r+="<tr><td>Free DRAM:</td><td>" + String(esp_get_free_internal_heap_size()) + " bytes</td></tr>\n"; 
   r+="</table><p><a href=\"https://github.com/HomeSpan/HomeSpan\">Click Here to Access HomeSpan Repo</a><p>";
```
frankonski, line 100: oh, I also just noticed the ending P tag is not closed....

HomeSpan:
I think you may have pulled down a copy of the code before I finished tweaking.  The current dev branch is "release-1.8.1-dev" and the callback passes a String& and returns void (which should matching the doc).

frankonski:
> I think you may have pulled down a copy of the code before I finished tweaking. The current dev branch is "release-1.8.1-dev" and the callback passes a String& and returns void (which should matching the doc).

I think so, because I cloned a "release-1.8.1" branch yesterday, which had both `library.properties` and `settings.h` showing `1.8.0`, and I was about to ask if a `-dev` could actually be set to a `1.8.1` release to avoid confusion.

HomeSpan:
I fixed the missing `</p>` in example shown in Logging.md.  Apart from that, can we close this PR?  Note the Settings.h file should already have 1.8.1 as the version number.  I usually don't update the library.properties version number until I'm ready to release (there are some things I still don't understand about the Arduino script that pulls library updates from GitHub and want to make sure it does not pull anything prematurely).

frankonski:
The new callback using a pointer to the String also works fine. But doc still requires two small edits: reference to the function name (&) and function name itself if you want it to match with the example code just above.

`homeSpan.setWebLogCallback(extraWebData)` should be `homeSpan.setWebLogCallback(&extraData)`

HomeSpan:
Thanks for catching this.  I updated the doc so that the Callback references `extraData` to match the actual example function.  However, I did not include the `&` since it is redundant for pointers to functions and its use is more a matter of style (I haven't used it other HomeSpan examples).

frankonski:
> However, I did not include the `&` since it is redundant for pointers to functions and its use is more a matter of style (I haven't used it other HomeSpan examples).

Well, after reading about what you just said, it looks like I just learned something new. It should be a good day today!

PR #672, opened 2023-10-31
Add hint for UTF8 usage for names

mgeramb:
- Improved example
- Hint in documentation


HomeSpan:
Is the `u8` prefix needed?   Seems like the compiler used in the Arduino IDE generates the same bytes (coded in UTF-8) with or without the `u8`.

mgeramb:
If you do not use the prefix, the encoding will depend on the encoding of the source file and/or compiler, so you don‘t have a guarantee for it.  https://stackoverflow.com/questions/331690/using-unicode-in-c-source-code

PR #679, opened 2023-11-08
Adding enum for SecuritySystemState values.

frankonski:
Ensures valid values are used, makes code easier to read and maintain.

### DIFF
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -378,6 +378,7 @@ This is a **base class** from which all HomeSpan Characteristics are derived, an
 * `void setVal(value [,boolean notify])`
   * sets the value of a numerical-based Characteristic to *value*, and, if *notify* is set to true, notifies all HomeKit Controllers of the change.  The *notify* flag is optional and will be set to true if not specified.  Setting the *notify* flag to false allows you to update a Characateristic without notifying any HomeKit Controllers, which is useful for Characteristics that HomeKit automatically adjusts (such as a countdown timer) but will be requested from the Accessory if the Home App closes and is then re-opened
   * works with any integer, boolean, or floating-based numerical *value*, though HomeSpan will convert *value* into the appropriate type for each Characteristic (e.g. calling `setValue(5.5)` on an integer-based Characteristic results in *value*=5)
+  * when setting a SecuritySystemState value, use the template form of `->setVal<int>(HSSecuritySystemState::Away)`
   * throws a runtime warning if *value* is outside of the min/max range for the Characteristic, where min/max is either the HAP default, or any new min/max range set via a prior call to `setRange()`
   * *value* is **not** restricted to being an increment of the step size; for example it is perfectly valid to call `setVal(43.5)` after calling `setRange(0,100,5)` on a floating-based Characteristic even though 43.5 does does not align with the step size specified.  The Home App will properly retain the value as 43.5, though it will round to the nearest step size increment (in this case 45) when used in a slider graphic (such as setting the temperature of a thermostat)
 
@@ -396,7 +397,7 @@ This is a **base class** from which all HomeSpan Characteristics are derived, an
   * works on Characteristics with UINT8, UINT16, UINT32, and INT formats only
     * a warning message is thrown, and the request is ignored, if this method is called on a Characteristic with any other format
   * returns a pointer to the Characteristic itself so that the method can be chained during instantiation
-  * example: `(new Characteristic::SecuritySystemTargetState())->setValidValues(3,0,1,3);` creates a new Valid Value list of length=3 containing the values 0, 1, and 3.  This has the effect of informing HomeKit that a SecuritySystemTargetState value of 2 (Night Arm) is not valid and should not be shown as a choice in the Home App
+  * example: `(new Characteristic::SecuritySystemTargetState())->setValidValues(3,Stay,Away,Disarmed);` creates a new Valid Value list of length=3 containing the values 0 (Stay), 1 (Away), and 3 (Disarmed).  This has the effect of informing HomeKit that a SecuritySystemTargetState value of 2 (Night) is not valid and should not be shown as a choice in the Home App
 
 #### The following methods are supported for string-based Characteristics (i.e. a null-terminated C-style array of characters):
 
--- a/docs/ServiceList.md
+++ b/docs/ServiceList.md
@@ -145,8 +145,8 @@ RotationDirection|int|0|[0,1]|
 RotationSpeed|double|0|[0,100]|
 Saturation|double|0|[0,100]|
 SecuritySystemAlarmType|uint8_t|0|[0,1]|
-SecuritySystemCurrentState|uint8_t|3|[0,4]|
-SecuritySystemTargetState|uint8_t|3|[0,3]|
+SecuritySystemCurrentState|HSSecuritySystemState|HSSecuritySystemState::Disarmed|[0,4]<br>HSSecuritySystemState::Stay<br>HSSecuritySystemState::Away<br>HSSecuritySystemState::Night<br>HSSecuritySystemState::Disarmed<br>HSSecuritySystemState::Alarm|
+SecuritySystemTargetState|HSSecuritySystemState|HSSecuritySystemState::Disarmed|[0,3]<br>HSSecuritySystemState::Stay<br>HSSecuritySystemState::Away<br>HSSecuritySystemState::Night<br>HSSecuritySystemState::Disarmed|
 SerialNumber|char \*|"HS-12345"|||
 ServiceLabelIndex|uint8_t|1|[1,255]|
 ServiceLabelNamespace|uint8_t|1|[0,1]|
--- a/src/HAPConstants.h
+++ b/src/HAPConstants.h
@@ -87,3 +87,12 @@ enum class StatusCode {
   InvalidValue=-70410,  
   TBD=-1                       // status To-Be-Determined (TBD) once service.update() called - internal use only
 };
+
+// Based on HomeKit documentation for HMCharacteristicTypeCurrentSecuritySystemState
+typedef enum {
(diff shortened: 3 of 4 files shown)


frankonski:
For HomeSpan users, this ends up being much more friendly to use and maintain, especially if Apple decides it changes their numerical values for states. Unlikely to happen, but who knows..

Going from:
```
    partitionCurrentState = new Characteristic::SecuritySystemCurrentState(3);  // Sets initial state to disarmed
    partitionCurrentState->setValidValues(3,0,1,3);
    partitionTargetState = new Characteristic::SecuritySystemTargetState(3);    // Sets initial state to disarmed
    partitionTargetState->setValidValues(3,0,1,3);

    switch (partitionCurrentState->getVal()) {
      case 0:
        str = "Armed for Stay";
        break;
      case 1:
        str = "Armed for Away";
        break;
      case 2:
        str = "Armed for Night";
        break;
      case 3:
        str = "Disarmed";
        break;
      case 4:
        str = "ALARM!";
        break;
      default:
        str = "Unknown state";
    }
```
to
```
    partitionCurrentState = new Characteristic::SecuritySystemCurrentState(HSSecuritySystemState::Disarmed);  // Sets initial state to disarmed
    partitionCurrentState->setValidValues(3,Stay,Away,Disarmed);
    partitionTargetState = new Characteristic::SecuritySystemTargetState(HSSecuritySystemState::Disarmed);    // Sets initial state to disarmed
    partitionTargetState->setValidValues(3,Stay,Away,Disarmed);

    switch (partitionCurrentState->getVal()) {
      case HSSecuritySystemState::Away:
        str = "Armed for Away";
        break;
      case HSSecuritySystemState::Night:
        str = "Armed for Night";
        break;
      case HSSecuritySystemState::Stay:
        str = "Armed for Stay";
        break;
      case HSSecuritySystemState::Disarmed:
        str = "Disarmed";
        break;
      case HSSecuritySystemState::Alarm:
        str = "ALARM!";
        break;
      default:
        str = "Unknown state";
    }
```

HomeSpan:
Though this is a good idea, I had planned on adding something more generic to cover all the Characteristics that have a set selection of values.  To make the nomenclature as easy as possible, I thought I'd add an `enum` to the instantiation of each Characteristic in the `Span.h` file.  This would mean the fully scoped enum would be something like Characteristic::SecuritySystemCurrentState.Night, or, in your example above, you could use partitionCurrentState.Night, since the enums would be inherited. 

Since I created a macro to instantiate each Characteristic, I need to first update it so it takes an optional set of "enums".   I was also going to add the descriptions of each enum directly in the code as a tagged comment that would allow the auto-generation of documentation (I already have a simple script that generates the Services table from the code).

Though the code change is modest, it's a reasonable amount of work to go through all the HAP Characteristics, add all the enums, and then the descriptions of each enum (in some cases I may change the HAP description when it is unclear).

Was planning on tackling this after the 1.8.1 release.  Once I update the macro to make it as easy as possible to add enums to Characteristics I'll let you know so you can try it out.

phybros:
+1 for this, it would be nice to have states for OPEN, CLOSED etc for garage doors also, instead of 0, 1 and so on

frankonski:
> Since I created a macro to instantiate each Characteristic, I need to first update it so it takes an optional set of "enums". I was also going to add the descriptions of each enum directly in the code as a tagged comment that would allow the auto-generation of documentation (I already have a simple script that generates the Services table from the code).

Yeah, I noticed you had mastered the art of macros. I was happily impressed when I saw that. I might give a try to your original idea. Note that I'd also like to have this macro create c-style comments (if at all possible) so that when you use VSCode, you have inline documentation. Quite useful.

frankonski:
How is this updated PR? If you like the new style I will update the other .md files modified as part of this PR to make sure when it gets merge (if it does), the updated .md matches the new enums.

For now, it would be used this way:
```
    switch (partitionCurrentState->getVal()) {
      case Characteristic::SecuritySystemState::Away:
        str = "Armed for Away";
        break;
      case Characteristic::SecuritySystemState::Night:
        str = "Armed for Night";
        break;
      case Characteristic::SecuritySystemState::Stay:
        str = "Armed for Stay";
        break;
      case Characteristic::SecuritySystemState::Disarmed:
        str = "Disarmed";
        break;
      default:
        str = "Unknown state";
    }
```

I also added _Leak Status_, _Battery Charging Status_ and _Door State_.

HomeSpan:
I have this section of the code open right now -- I found a way of significantly reducing the memory associated with how the list of required and optional Characteristics for each Service are stored.  Please hold off a bit on any further changes.

frankonski:
> Please hold off a bit on any further changes.
sure can do! Should I just close that PR then?

HomeSpan:
Leave the PR open for now so we'll know to revisit soon.

HomeSpan:
I've updated the 1.8.1-dev branch with enums for all applicable Characteristics.  See *Span.h* for the changes I made to the macro and the list of all constants.

Constants are scoped within each Characteristic.  As such, there are two ways of accessing:

1. provide the fully-qualified name, such as `Characteristic::SecuritySystemTargetState::ARM_STAY`, or
2. if you create a Characteristic of any given type, the constants are automatically inherited. 

Example:

```C++
Characteristic::SecuritySystemTargetState target();
...
if(target.getNewVal()==target.ARM_STAY) {...}   // applicable constants are inherited by instances of a specific Characteristic 
```

This is how it works with pointers:

```C++
Characteristic::SecuritySystemTargetState *target;
...
target=new Characteristic::SecuritySystemTargetState();
...
if(target->getNewVal()==target->ARM_STAY) {...}
```

Note this will not work:

```C++
SpanCharacteristic *target;
...
target=new Characteristic::SecuritySystemTargetState();
...
if(target->getNewVal()==target->ARM_STAY) {...}     // <- fails because target was defined as general SpanCharacteristic
```

But if you use the above construct (as I have done in all the HomeSpan Examples ☹️) you can always use the fully-qualified name:

```C++
SpanCharacteristic *target;
...
target=new Characteristic::SecuritySystemTargetState();
...
if(target->getNewVal()==Characteristic::SecuritySystemTargetState::ARM_STAY) {...}     // always will work
```

I'll probably go back and update all the examples to avoid the generic use of SpanCharacteristic and add in all the constants wherever applicable.

Please feel free to give it a try and let me now how it works.

frankonski:
> Re the constants, one of the tricky part is to figure out a good naming convention, so curious if you find the names easy-to-use.

I think so, yes. I would still add enums for boolean characteristics (MotionDetect) as someone doing a `getVal()` on a variable name called `motionSensor` may want to use `DETECTED`.

And a suggestion: maybe `BATTERY_OK` is better than `NOT_LOW_BATTERY` ?

frankonski:
actually, it won't compile weather when using named or unnamed enum.
```
C:\Users\<hidden>\Documents\Arduino\libraries\HomeSpan\src/HomeSpan.h:613:13: error: invalid conversion from 'int' to 'Characteristic::ContactSensorState::<unnamed enum>' [-fpermissive]
     return(0);       // included to prevent compiler warnings
     
C:\Users\<hidden>\Documents\Arduino\libraries\HomeSpan\src/HomeSpan.h:613:13: error: invalid conversion from 'int' to 'Characteristic::ContactSensorState::state' [-fpermissive]
     return(0);       // included to prevent compiler warnings
```
The fix seems to be to change
```
    return (0);       // included to prevent compiler warnings  
```
to
```
    return T();
```
.. but don't ask me why! ;-)

frankonski:
Maybe this would sound better?
`contact.NO_CONTACT` and `contact.DETECTED`,
`motion.NO_MOTION` and `motion.DETECTED`

instead of `NOT_DETECTED` ?

HomeSpan:
Did this compiler warning just start?  The return(0) has been there for quite a while.

frankonski:
> Did this compiler warning just start? The return(0) has been there for quite a while.

Yes, but I think it showed up only once I started actually using the enum. And it wasn't a warning, it was actually an error preventing the build to complete.
Here is the full error:
```
In file included from C:\Users\<hidden>\Documents\<hidden>\dscHome\dscHome.ino:59:
C:\Users\<hidden>\Documents\Arduino\libraries_renamed\HomeSpan\src/HomeSpan.h: In instantiation of 'T SpanCharacteristic::uvGet(SpanCharacteristic::UVal&) [with T = Characteristic::SecuritySystemTargetState::<unnamed enum>]':
C:\Users\<hidden>\Documents\Arduino\libraries_renamed\HomeSpan\src/HomeSpan.h:774:12:   required from 'void SpanCharacteristic::setVal(T, boolean) [with T = Characteristic::SecuritySystemTargetState::<unnamed enum>; boolean = bool]'
C:\Users\<hidden>\Documents\<hidden>\dscHome\DEV_dscPartition.h:181:69:   required from here
C:\Users\<hidden>\Documents\Arduino\libraries_renamed\HomeSpan\src/HomeSpan.h:613:14: error: invalid conversion from 'int' to 'Characteristic::SecuritySystemTargetState::<unnamed enum>' [-fpermissive]
     return (0);       // included to prevent compiler warnings
```

HomeSpan:
I've been able to reproduce the issue.  It's actually a real problem I need to address as a result of implicit typecasting when using templates that have enum constants.  The line `return(0)` should never be reachable, but I now see why it is when called with the new enum constants.  I'll address and post a fix once I figure out the best solution.

HomeSpan:
I figured out the problem and it turns out the solution is to indeed change `return(0)` to `return((T)0)`.  This line never does get called, but the compiler will complain anyway. Fix is uploaded to 1.8.1-dev.

frankonski:
Compiles fine, runs fine for now with that last fix, and enums seems to be enum-ing just fine. Thanks!

frankonski:
I'm happy with your changes, feel free to close that PR.

PR #685, opened 2023-11-11
Add temp buffer statistic

mgeramb:
I have added a function to get information about the memory usage of the temp buffer

### DIFF
--- a/src/HAP.cpp
+++ b/src/HAP.cpp
@@ -96,7 +96,7 @@ void HAPClient::init(){
   }
 
   if(!nvs_get_blob(hapNVS,"CONTROLLERS",NULL,&len)){                // if found long-term Controller Pairings data from NVS
-    TempBuffer <Controller> tBuf(len/sizeof(Controller));
+    TempBuffer <Controller> tBuf(len/sizeof(Controller), "HAPClient::init");
     nvs_get_blob(hapNVS,"CONTROLLERS",tBuf.get(),&len);             // retrieve data
     for(int i=0;i<tBuf.size();i++){
       if(tBuf.get()[i].allocated)
@@ -157,7 +157,7 @@ void HAPClient::processRequest(){
     return;
   }
  
-  TempBuffer <uint8_t> httpBuf(messageSize+1);      // leave room for null character added below
+  TempBuffer <uint8_t> httpBuf(messageSize+1, "HAPClient::processRequest");      // leave room for null character added below
   
   if(cPair){                           // expecting encrypted message
     LOG2("<<<< #### ");
@@ -859,7 +859,7 @@ int HAPClient::getAccessoriesURL(){
   LOG1(")...\n");
 
   int nBytes = homeSpan.sprintfAttributes(NULL);        // get size of HAP attributes JSON
-  TempBuffer <char> jBuf(nBytes+1);
+  TempBuffer <char> jBuf(nBytes+1,"HAPClient::getAccessoriesURL");
   homeSpan.sprintfAttributes(jBuf.get());                  // create JSON database (will need to re-cast to uint8_t* below)
 
   char *body;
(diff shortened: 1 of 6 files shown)


HomeSpan:
I'm hesitant to add this to the production branch of HomeSpan since the user really can't control the TempBuffers.  However, I can see the use while trying to develop or change the actual HomeSpan library (as opposed to end-users who are developing sketches *with* the library).  In my own work, I usually add lots of Serial.printf statements so I can see exactly when buffers are being called, along with any other diagnostics that help describe the circumstances during which a buffer was used.  Then I delete all these Serial.printf statement once satisfied everything is working.

There is definitely more room to optimize memory.  If you look at a branch I called "server" you'll see a number of other optimizations I made leading up to, and beyond those in 1.8.1:

1) Rather than keep a fixed-size buffer that can store up to 16 HAP connections, I completely refactored the connections code to use dynamic storage that is created and destroyed as HAP socket connections are opened and closed.  Prior to iOS 14, HomeKit devices needed to support many simultaneous HAP connections since each controller communicated directly with each HomeKit device.  In other words, if you had 3 iPhones, 2 HomePods, and 3 iMacs, HomeKit would often open and retain 8 simultaneous connections.  However, with the intro of using a HomePod or Apple TV as a Home Hub, HomeKit routes all requests through a single device, and rarely needs to open more than 2 simultaneous connections, which means a fixed buffer supporting 16 HAP connections is a waste of space.  Changing this was quite intrusive and I needed to completely re-work other parts of the code as well (such as Event Notifications) that were tied to HAP Connections.  I did get it working but during testing I noticed some instabilities that may have been because of the code changes, or may have been a result of other HomeKit changes and updates that were occurring around the same time.  Hence the 1.8.1 branch commits stop just before this code change, but I do intend to revisit once 1.8.1 is released since I really liked how this streamlined connection management.

2) TLV data - These structures are mostly used for pairing and pair-verify calls.  HAP defines 11 different TLV types, and I reserve space for all of them during start-up.  Though it's not a huge amount (about 1.5K), I refactored the code so that space is allocated only when needed, and destroyed after.  

3) SRP data and constants - These structure are used for pairing only, and involve manipulating arbitrary-sized numbers using the MBED TLS Library (which is included into the Espressif IDF).  Some of these numbers can get quite large (taking up 700+ bytes), though the total size used across all data is probably only 3K. Nevertheless, there is a bunch optimization (and a lot of refactoring) I was able to due to keep memory use to a minimum, as well as to deallocate all these structures once pairing is completed.

Since I made the changes (2) and (3) above after (1) in the server branch, these changes are not in 1.8.1.  I may revisit (2) and (3) at some point in the future, but it's a lot of changes for not that much gain.

Other changes in the server branch are related to my goal of refactoring HomeSpan into two components - a server and a wrapper that calls the server.  HomeSpan would continue to use both, so the changes would be hidden from the typical end user.  However, it would be possible to just use the server part without the wrapper.  The server would contain nothing but core HAP functionality so the user can implement their own WiFi connectivity, their own web logging, etc., or they could pick and choose any combinations of HomeSpan functionality, such as including SpanPoint but excluding SpanButton.

In principle this would yield a smaller overall footprint in terms of code size as well as memory usage.  You'll see some of the changes I made in the server branch to separate out "optional" features from core HAP functionality.  Unfortunately, this proved to be more difficult than worthwhile because I am trying to do all of this from within the Arduino IDE, which provides very little control of how libraries are compiled in real time.  If this was just a standalone library to be used in a normal C++ Unix environment, there are lots of ways of ensuring only those components needed are called.  But that's much more difficult in the Arduino IDE.

I think the solution is to use the power of the Arduino IDE to my advantage, instead of fighting it.  If every function I wanted to make optional was in a separate *library*, users of the server version would be able to add those functions by simple inclusions of *.h files.  However, if I go this route, HomeSpan could not be distributed as a single library through the Arduino Library Manager.  It would only be "installable" from a zip file via GitHub.  

Lots to consider for the future...

mgeramb:
Thank you for your long explanation. I will close this PR and continue with the work of my context object. I will remove the diagnostic feature at the end.

PR #686, opened 2023-11-12
Send encryption context (Draft)

mgeramb:
This is the first version of using a context object to minimize the heap buffer size. I want discuss the basic concept with you with this draft PR.

Basically it works like that:

-  New abstract class `CallContext` collect string data in a buffer
-  If the buffer is full a virtual function `handlePage` is called
-  A new class `SendEncryptedCallContext` derived from `CallContext` encrypts the content and send it to the network
-  A new class `CalcHashContext` derived from CallContext calculates a hash out of the string

The prototype usage is implemented in the `HAPClient::getAccessoriesURL()` and `Span::updateDatabase`. The function `Span::sprintfAttributes()` split now the memory blocks for each accessory.

My next step would be to use the context object every where and simple add the string parts to the context object with the `printf` and `print` function, the function to reserve a string memory block can be made private after this change. So no manual memory buffer allocation would be necessary. 

Also the counting of all required bytes for the `Content-Length` header could be optimized by using `chunked` transfer encoding. But then, the context have to had knowledge about HTTP protocol, thats the reason why I'am not yet happy with this idea.

The memory usage is already nearly perfect, the largest TempBuffer size is 1120. No dependency of the number of accessories anymore.

The screenshot is my application with 50 mixed device types:
image

Please let me know what are you thinking about the basic concept and the current implementation.

### DIFF
--- a/src/CallContext.cpp
+++ b/src/CallContext.cpp
@@ -33,7 +33,7 @@
 #include <sodium.h>
 
 CallContext::CallContext()
-    : stringBuffer(1024, "CallContext stringBuffer") 
+    : stringBuffer(MINIMUM_STRING_BUFFER_SIZE, "CallContext stringBuffer") 
 {
 }
 
@@ -47,7 +47,6 @@ void CallContext::printf(const char* format, ...)
     LOG2(buffer);
     va_end(ap);
 }
-
 
 char* CallContext::reserveStringBuffer(int stringLength)
 {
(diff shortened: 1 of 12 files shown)


HomeSpan:
This looks quite extensive - thanks for all the work.  It will take me some time to review (may have to wait for the next weekend), but very promising.

mgeramb:
I have tested now with your latest version of 1.8.1-dev. There is no big difference anymore in memory consumption and it seems the the high stack usage come from my solution. I close the the PR for this reason.

PR #691, opened 2023-11-16
Saving 10 bytes twice

frankonski:
HAP config number can't be higher than 65535, which fits into 6 bytes rather than 16, including null.

### DIFF
--- a/src/HomeSpan.cpp
+++ b/src/HomeSpan.cpp
@@ -482,7 +482,7 @@ void Span::checkConnect(){
 
   // add MDNS (Bonjour) TXT records for configurable as well as fixed values (HAP Table 6-7)
 
-  char cNum[16];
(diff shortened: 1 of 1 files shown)


HomeSpan:
These are just small temp spaces.  They pull temp memory from stack and return after the end of the function, so not an issue if a few extra bytes are used.  When debugging I often temporarily add more diagnostics to a sprintf function, which requires a bit more memory.  Adding some extra memory means I don't need to keep changing the size of the buffer, so I usually leave them as is once I finish coding a section.

Heap buffers are different - this is where new objects are stored so it's always good to minimize the amount you need since storage is cumulative across functions and scopes (unlike the stack).

frankonski:
cool, no worries. Closing then. There I was hoping to safe few bytes ;-)

Issue #645, opened 2023-09-15
Using .setControlPin() faults pairing


gotfredsen:
When I use `homeSpan.setControlPin(3);` in my code, the button works for factory resetting, but then I can't pair. The serial console says the same in both cases:

> DEVICE NOT YET PAIRED -- PLEASE PAIR WITH HOMEKIT APP

but the LED (in my case a Neopixel) shows a long series of fast blink, when I use `homeSpan.setControlPin(3);` ... and the usual double blink when I have commented out `//homeSpan.setControlPin(3);`, allowing me to pair.

Below is the schematic for the button, in case it could be hardware, but as I mentioned, pressing the button does initiates a factory reset - `BUTTON` goes straight to GPIO3 (physically pin 8) on the ESP32-C3 (specifically an ESP32-C3FH4 w/built flash):
![button_schematic](https://github.com/HomeSpan/HomeSpan/assets/2776277/7c8573f8-f860-4b26-9f3b-c14955de5314)

I have tried all combination of `pinMode(3, OUTPUT or INPUT or INPUT_PULLUP);` prior to `homeSpan.begin()`, to see if that would help, but no luck.

Any suggestions would be appreciated.

XDM525:
I believe it may have to do with your button setup. What I would try is remove the line of code pinMode(3, OUTPUT or INPUT or INPUT_PULLUP); as I'm pretty sure homeSpan.setControlPin(3); sets the control pin as INPUT_PULLUP. Then take pin1 from your switch, remove it from VCC and attach it directly to GPIO3 and remove R12. This is what I'm doing and it works for me.

HomeSpan:
That's correct.  The Control pin should be connected to GROUND via a Pushbutton, not VCC.  And HomeSpan configures the  pin automatically.

gotfredsen:
Thank you all.

Like there is a `setStatusDevice` as an alternative to `setStatusPin`, is there a  `ControlDevice` as an alternative to `setControlPin`, which would help me here?

Also can you use CapTouch as a `ControlPin`? I know with `SpanButton` you can do a `TRIGGER_ON_TOUCH`, but can't see if that is available with `setControlPin`....

HomeSpan:
Sorry, at present the control pin is hardcoded to use a pushbutton connected to ground.

gotfredsen:
Ok

stzoran1:
I think I know how we can enhance setControlPin() method to allow push buttons which are pulled down.

[...]

PR #701, opened 2023-11-23
Added support for pull down control buttons

stzoran1:
See #645

### DIFF
--- a/src/HomeSpan.h
+++ b/src/HomeSpan.h
@@ -292,6 +292,7 @@ class Span{
   boolean deleteAccessory(uint32_t aid);             // deletes Accessory with matching aid; returns true if found, else returns false 
 
   void setControlPin(uint8_t pin){controlButton=new PushButton(pin);}     // sets Control Pin   
+  void setControlPinHigh(uint8_t pin) { controlButton = new PushButton(pin, PushButton::TRIGGER_ON_HIGH); } // sets Control Pin for pulled down push buttons
   void setStatusPin(uint8_t pin){statusDevice=new GenericLED(pin);}       // sets Status Device to a simple LED on specified pin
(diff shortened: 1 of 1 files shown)


stzoran1:
If you like this new feature I can update docs before merge. Please let me know.

HomeSpan:
Thanks, but I think it would be more generic to simply add a second optional argument to setControlButton to allow the user to specify the Trigger Type, which could be one of the built-ins, or a complete custom function:

```C++
setControlPin(uint8_t pin, triggerType_t triggerType=TRIGGER_ON_LOW){controlButton=new PushButton(pin, triggerType_t);}
```

I can add this to the next release (in progress at release-1.8.1-dev)

stzoran1:
I completely agree @HomeSpan

HomeSpan:
Completed.  

`homeSpan.setControlPin()` now accepts a *second* optional argument specifying the trigger type using the same logic as SpanButton.  Choices are:

SpanButton::TRIGGER_ON_LOW --- for control pin connected to ground through a pushbutton (the default)
SpanButton::TRIGGER_ON_HIGH --- for control pin connected to VCC through a pushbutton
SpanButton::TRIGGER_ON_TOUCH --- for a control pin connected to a touch sensor (not really recommended for this use)

or you can provide your own trigger function of form  `boolean f(int pin)`  For example, using a lambda function:

```
pinMode(21,OUTPUT);
homeSpan.setControlPin(21, [](int pin)->boolean{ return(digitalRead(pin)); } );
```

is equivalent to:

```
homeSpan.setControlPin(21, SpanButton::TRIGGER_ON_HIGH);
```

Note as per above, if you specify your own trigger function, you need to explicitly configure the pin (or whatever device you are using for the control pin).

Code is updated in 1.8.1-dev branch and will be part of next production release (with documentation updated accordingly).

Thanks again for the suggestion.

PR #717, opened 2023-12-19
Support use as ESP IDF component

### LINKED ISSUE #708: ESP-IDF integration
Hello it's an issue/question, I would like to use HomeSpan library in VScode ESP-IDF extension I have added arduino as component and run across multiple problems. First was no "core_version.h" file I solved that, then terminal complains about undetected files in src/extras/Pixel.h:139:30, no sodium.h file etc. I was wondering if there is an easy way to use this library outside Arduino ?

dzungpv:
Fix #708 Support use as ESP IDF component.
How to use: https://docs.espressif.com/projects/arduino-esp32/en/latest/esp-idf_component.html


dzungpv:
@HomeSpan  why you don't merge this?

HomeSpan:
I have open code with significant updates in progress.   Will be a while before I get to this (also, I'm not using the IDF environment but it's been on my list to try this out, so appreciate the PR).    Will let you know when I begin and may have some questions.

PR #725, opened 2023-12-27
Add function to read out the stack usage in the autopoll function

mgeramb:
It would be nice to have a possibility to get the stack usages of the autopoll function. Another option would be to return the taskhandle from the autoPoll() function.


HomeSpan:
Thanks for the suggestion.  I decided to make it generic and added `Taskhandle_t homeSpan.getAutoPollTask()` to return the handle to the autoPoll task (or NULL if autoPoll is not used).  I also added an extra line to the 'm' command that lists the lowest stack levels for both the Arduino loop() task and the autoPoll() task (unless autoPoll is not used).

Code has been updated in 1.8.1-dev.

PR #730, opened 2024-01-03
Adding TLS support for weblogs

frankonski:
This is a [draft pull request](https://github.blog/2019-02-14-introducing-draft-pull-requests/) btw..

Logs can contain sensitive data.
When logs are visible via the network (HTTP), logs fly in clear text on the wire. Not good.

This is a first stab at getting some feedback from @HomeSpan maintainer. I think it is somewhat very easy to use, if you look at the updated `Logging.md`.

There are still some unknowns I would like you feedback on. Look at the **TODO** comment block in `generate_certificate()` of `HomeSpan.cpp`.

The idea is that HomeSpan users should provide their own certificate & private key. Until they do, HomeSpan will generate a new keypair/certificate at each reboot. A bit annoying for developers and browser security warnings, but this is on purpose to force them to use their own. Should they decide not to use their own certificates, then HomeSpan still keep weblogs secure by not reusing the same keypair for multiple devices.

Only ECDSA is supported because keys are smaller than RSA while providing the same level of protection.

Awaiting feedback!

### DIFF
--- a/docs/Logging.md
+++ b/docs/Logging.md
@@ -2,6 +2,13 @@
 
 HomeSpan includes a variety of message logs with different levels of verbosity, as well built-in methods to create your own log messages and web logs.
 
+1. [HomeSpan Log Message](homespan-log-messages)
+1. [User-Defined Log Messages](user-defined-log-messages)
+1. [Web Logging](web-logging)
+   1. [Custom Style Sheets (CSS)](custom-style-sheets-css)
+   1. [Adding User-Defined Data and/or Custom HTML](adding-user-defined-data-and-or-custom-html)
+   1. [Enabling TLS Web Logs](enabling-tls-web-logs)
+
 ## HomeSpan Log Messages
 
 HomeSpan log messages are typically output directly to the Arduino Serial Monitor with three possible levels of verbosity:
@@ -104,6 +111,15 @@ To embed this custom HTML text in the Web Log, call `homeSpan.setWebLogCallback(
 
 Note that *r* is being passed as a reference and already includes all the HTML text the Web Log has produced to set up the page and create the initial table.  You should therefore *extend* this String using  `+=`.  Do not simple assign this variable to a new String with `=` or you will overwrite all the HTML already produced!
  
+### Enabling TLS Web Logs
+When Web Logs are enabled, HomeSpan shares the same web server instance to serve HomeKit controllers and web logs. Web logs are accessed via HTTP only, providing no security. If sensitive information is expected to be logged, web logs should be accessed via HTTPS, using Transport Security Layer (TLS).
+
+To enable TLS with default configuration, call `enableTLS()` prior to calling `homeSpan.begin()`. HomeSpan will then attempt to generate an ECDSA keypair and self-signed certificate using NIST P-256 curve. Since the RSA algorithm requires larger keys, the use of Elliptic Curve Cryptography (ECC) is preferred as it uses smaller key sizes while providing the same security strength. RSA keypairs are certificates are not currently supported.
+
+A new EC keypair and ECDSA certificate will be created after each boot unless developers want to use their own, using `enableTLS(port, ec_private_key_pem, ec_server_cert_pem)`. This ensures that no HomeSpan device ends up with the same keypair which would be considered a security vulnerability. HomeSpan developers should generate their own keypair and certificate, PEM-encoded, and pass their values to `enableTLS(port, ec_private_key_pem, ec_server_cert_pem)`. **Never log the private key.**
+
+When TLS is configured and web logs are accessed via the non-HTTPS port, HomeSpan will issue an HTTP 301 (Permanently Moved) to the HTTPS port, which defaults to 443.
+
 ---
 
 [↩️](../README.md) Back to the Welcome page
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -200,6 +200,17 @@ The following **optional** `homeSpan` methods enable additional features and pro
   * returns the version of a HomeSpan sketch, as set using `void setSketchVersion(const char *sVer)`, or "n/a" if not set
   * can by called from anywhere in a sketch
 
+* `Span& enableTLS()`
+  * Enable TLS (HTTPS) for weblogs, instead of unsecure (non-TLS / HTTP) connections. If weblogs have been enabled for a specific URI, end users will be redirected to the HTTPS port.
+  * Defaults to using port 443 (default HTTPS port)
+  * New ECDSA certificate generated at each reboot
+* `Span& enableTLS(u_int16_t port = 443, const char *ec_private_key_pem = NULL, const char *ec_server_cert_pem = NULL)`
+  * Enable TLS (HTTPS) for weblogs, instead of unsecure (non-TLS / HTTP) connections. If weblogs have been enabled for a specific URI, end users will be redirected to the HTTPS port.
+  * **port** Port for HTTPS connections. Defaults to 443 (default HTTPS port)
+   * **ec_private_key_pem** ECC private key, in PEM format, to use for HTTPS. NOTE: RSA keys are not supported given their larger key size for similar security strength. A 256-bit ECC key provides similar security strength as an RSA 4096-bit key. When NULL, HomeSpan will attempt to generate its own ECC keypair at EACH REBOOT.
+   * **ec_server_cert_pem** Matching EC certificate for the primate key. Keypair validation is not done.  When NULL, HomeSpan will attempt to generate its own ECDSA with NIST P-256 certificate at EACH REBOOT.
+   
+
 * `Span& enableWebLog(uint16_t maxEntries, const char *timeServerURL, const char *timeZone, const char *logURL)`
   * enables a rolling Web Log that displays the most recent *maxEntries* entries created by the user with the `WEBLOG()` macro.  Parameters, and their default values if unspecified, are as follows:
     * *maxEntries* - maximum number of (most recent) entries to save.  If unspecified, defaults to 0, in which case the Web Log will only display status without any log entries
--- a/src/HAP.cpp
+++ b/src/HAP.cpp
@@ -305,7 +305,43 @@ void HAPClient::processRequest(){
     }
 
     if(homeSpan.webLog.isEnabled && !strncmp(body,homeSpan.webLog.statusURL.c_str(),homeSpan.webLog.statusURL.length())){       // GET STATUS - AN OPTIONAL, NON-HAP-R2 FEATURE
-      getStatusURL();
+
+      // if TLS server is not configured, business as usual, else redirect to it
+      if (!homeSpan.webLog.tls_server) {
+        getStatusURL();
+
+      } else {
+        char *line = strtok(body, "\n");
+        char *host = NULL;
+        while (line != NULL && host == NULL) {
+          if ((host = strstr(line, "Host:")) != NULL) {
+            host += 5; // Trim "Host:"
+            // Left-trim any white spaces the browser may put between "Host:" and the actual host
+            while(host[0] == ' ') host++;
+            // Trim the port number, if specified, as we have to redirect to a different port
+            for(int x=0; x<strlen(host); x++) {
+              switch(host[x]) {
+                case ':':
+                case '\r':
+                case '\n':
+                  host[x] = '\0';
+              }
+            } 
+          }
+          line = strtok(NULL, "\n");
+        }
+
+        // Send back an HTTP 301 redirect to the TLS server host (and port, if different than 443)
+        String resp = "HTTP/1.1 301 Moved Permanently\n";
+        resp += "Location: https://" + (host != NULL ? host : client.localIP().toString());
+        resp += homeSpan.webLog.tls_server_conf.port_secure != 443 ? ":" + String(homeSpan.webLog.tls_server_conf.port_secure) : "";
+        resp += String(homeSpan.webLog.statusURL.c_str() + 4) + "\n"; // statusURL contains "GET /status", so trim first 4 chars
+        LOG2("-------- HTTP Redirect -------\n%s\n---------------------\n", resp.c_str());
+        client.print(resp);
+        client.stop();
+
+      }
+
       return;
     }    
 
@@ -1189,7 +1225,7 @@ int HAPClient::putPrepareURL(char *json){
 
 //////////////////////////////////////
 
-int HAPClient::getStatusURL(){
+int HAPClient::getStatusURL(httpd_req_t *req){
 
   char clocktime[33];
 
@@ -1210,7 +1246,8 @@ int HAPClient::getStatusURL(){
     
   sprintf(uptime,"%d:%02d:%02d:%02d",days,hours,mins,secs);
 
-  String response="HTTP/1.1 200 OK\r\nContent-type: text/html; charset=utf-8\r\n\r\n";
+  // if req is set, HTTP response code and content type is set by httpd_resp_* APIs
+  String response=(req?"":"HTTP/1.1 200 OK\r\nContent-type: text/html; charset=utf-8\r\n\r\n");
 
   response+="<html><head><title>" + String(homeSpan.displayName) + "</title>\n";
   response+="<style>body {background-color:lightblue;} th, td {padding-right: 10px; padding-left: 10px; border:1px solid black;}" + homeSpan.webLog.css + "</style></head>\n";
@@ -1312,7 +1349,11 @@ int HAPClient::getStatusURL(){
 
       if(response.length()>1024){     // if response grows too big, transmit chunk and reset
         LOG2(response);
-        client.print(response);
+        if (req) {
+          httpd_resp_send_chunk(req, response.c_str(), HTTPD_RESP_USE_STRLEN);
+        } else {
+          client.print(response);
+        }
         delay(1);                     // slight pause seems to be required
         response.clear();
       }
@@ -1324,7 +1365,12 @@ int HAPClient::getStatusURL(){
 
   LOG2(response);
   LOG2("\n");
-  client.print(response);
+  if (req) {
+    httpd_resp_send_chunk(req, response.c_str(), HTTPD_RESP_USE_STRLEN);
+    httpd_resp_send_chunk(req, NULL, 0); // Required to complete the response.
+  } else {
+    client.print(response);
+  }
   LOG2("------------ SENT! --------------\n");
   
   delay(1);
--- a/src/HAP.h
+++ b/src/HAP.h
@@ -126,7 +126,7 @@ struct HAPClient {
   int getCharacteristicsURL(char *urlBuf);     // GET /characteristics (HAP Section 6.7.4)  
   int putCharacteristicsURL(char *json);       // PUT /characteristics (HAP Section 6.7.2)
   int putPrepareURL(char *json);               // PUT /prepare (HAP Section 6.7.2.4)
-  int getStatusURL();                          // GET / status (an optional, non-HAP feature)
+  int getStatusURL(httpd_req_t *req = NULL); // GET / status (an optional, non-HAP feature)
 
   void tlvRespond();                                                // respond to client with HTTP OK header and all defined TLV data records (those with length>0)
   void sendEncrypted(char *body, uint8_t *dataBuf, int dataLen);    // send client complete ChaCha20-Poly1305 encrypted HTTP mesage comprising a null-terminated 'body' and 'dataBuf' with 'dataLen' bytes
--- a/src/HomeSpan.cpp
+++ b/src/HomeSpan.cpp
@@ -154,7 +154,19 @@ void Span::begin(Category catID, const char *displayName, const char *hostNameBa
     delay(100);
     esp_ota_mark_app_invalid_rollback_and_reboot();
   }
-  
+
+  // If a private key has been set we can assume that TLS server is configured. TLS web server should be
+  // started/stopped when Wifi connects / disconnects
+  if (webLog.tls_server_conf.prvtkey_len > 0) {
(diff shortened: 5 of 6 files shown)


HomeSpan:
Some initial thoughts:

I think it would make more sense to add TLS into a sketch as a separate web server rather than try to integrate into HomeSpan for both design and practical reasons.

Regarding design, HomeSpan is not a web server.  It uses HTTP, but the server is completely custom as you can see from the code.  HomeSpan does not use any of the IDF http handlers or components.  The WebLog "server" is a convenient adjunct that was easy to add, but it's not designed to be a full web server.  In fact, I have a number of examples, such as "Programmable Hub" that do use ESP32 standard HTTP event handling, but that's done in the sketch - not within the HomeSpan library.

If I was already using ESP-IDF HTTP handles within HomeSpan, then adding in a TLS layer might make more sense, since you would be extending existing functionality.  But to use HTTPS, you are basically bringing in the entire IDF HTTP package - just for the web logging.  This seems too heavy to add for such a tenuous link to the library itself (just the web logging, nothing to do with HomeKit itself).

On the practical side, if you look at the latest commits, you'll see that the entire Web Log output has been re-written.  Instead of using Strings to create a buffer that is then transmitted, I am now using a customized std:ostream with on-the-fly encryption (called hapOut).  The motivation for this change has nothing to do with the Web Log, but rather I made this change for all of the HomeKit HAP transmissions.  I'm still benchmarking, but the reduction of memory-usage when using this new method is very signifiant and it is now possible to run a lot more Accessories in HomeSpan before running out of memory (not an issue of course if you are using PSRAM).  But because the Web Log is an adjunct to everything in HomeKit, since I made very substantive changes to the HomeKit HAP transmissions, I needed to update the Web Log stuff (which benefited Web Log memory usage in any case).

In others words, if you integrate a completely new server into HomeSpan, it's going to be unsustainable in the long term.  You are going to likely want to continue making changes, fixing issues, adding new functionality, etc, and I'm going to have to worry that any changes I make could break the HTTP handlers and TLS, which may only impact a a small number of users, but nevertheless I really try to make sure none of my change break anything others have done.

My recommendation would be to deploy your server within your sketch and then PULL IN the web log data to a callback you have that uses the http_chunk command.  This would be analogous to the hook I added a few months ago that lets you add your own HTML code to existing web logs by extending a String.  As noted above, I no longer use Strings in the Web Log, but you'll see that the hook is still there ---my changes should have been transparent.

I'd be happy to do the same for Web Log output.  I can imagine a simple HomeSpan command that you invoke with a callback function as an argument whenever you want to grab all the Web Log output.  You can then process the output in whatever way makes the most sense for your sketch (likely by calling http_chunk upon receiving each chunk from HomeSpan).   I can also add an option to start the HomeSpan Web Log as usual, but without enabling the actual HTTP get, so that the log can only be accessed through your code as HTTPS.

I would integrate this callback logic into the new hapOut streaming code just deployed so it would always track with the latest HomeSpan version and changes would be transparent to your sketch.

frankonski:
Hi, allow me to continue my sales pitch ;-)

> it would make more sense to add TLS [...] as a separate web server

No choice to have it as a separate web server. HomeSpan serves "plain encrypted text", over a non-TLS socket, over HTTP. To have another socket provide TLS (rather than clear text), HomeSpan would need a second "listening socket" (a.k.a, a second web server).

> HomeSpan is not a web server.
I'd disagree on that definition, since HomeSpan serves HTTP requests (`GET /accessories`, `PUT /prepare`) which are HAP-specific URI as well as `GET /status` which is non-HAP-specific that can be modified by HS users at their own will to return what they want. I'd say that you have implemented your own "web server", lightweight to serve HAP-specific URIs, etc. 

Could you have used an instance of `httpd_handle_t` to serve HAP URIs? Sure, but at what memory-consumption-cost? That I can't say, but probably higher memory footprint than how you currently do it. One would have to redo your HAP code to use `httpd_handle_t` instead and see the difference.. But that is a separate topic.

> but it's not designed to be a full web server.

hum... The world of web servers is a complicated one (Apache, ngnix, etc)... Do you know the biggest commercial usage of HomeSpan, and what would happen if your current implementation does not do standard security vulnerabilily mitigations? I know there is **Home** in HomeSpan, and the majority of users probably don't care about HomeSpan crashing, or halting, because they can unplug their device, plug it back in, and move on. But if HomeSpan is used in commercial products that can't and should not be DDoS'ed, then maybe thinking about using `httpd_handle_t` might be an easy mitigation. But again, a separate topic.

> But to use HTTPS, you are basically bringing in the entire IDF HTTP package - just for the web logging.

What if all TLS-specific functionalities are within `#ifdef` statements that gets defined (compiled in) when `HomeSpanTLSWebLog.h` is included? I was already thinking about submitting my initial PR this way, but thought I'd share my plans first with you.

> I am now using a customized std:ostream with on-the-fly encryption (called hapOut).

Yeah, I saw you mentioning something along these lines in another issue. Seems like good stuff. Side note, probably unrelated. I had interim code (instead of using `strtok`) that was using `std:isstream` that seemed a-ok, however that ended up filling up the .txt section of the binary, so compilation was failing. The simple fact of removing the #include statement solved the compilation issue. So I don't know if `std:ostream` would do the same. Will see..

I think the _loging callback_ approach is the best one. Being a library, HomeSpan may want to provide logs to its users. But meeting all users needs is impossible, That's why if HomeSpan would **only** provide a logging callback mechanism to log anything, that would allow users to do whatever they want with these logs.
Some may want to send logs to MQTT, some to a web server (like WebLog is currently doing), some to an HTTPS server (like I would prefer), some to an LCD screen with buffers... etc. Possibilities become infinite.

HomeSpan:
Some addition thoughts:

I'm not concerned about commercial use since by definition and Apple licensing, HomeSpan can only be used by hobbyists for their own personal devices.  They even remind you of this whenever you pair a new accessory.  I've looked into making a commercial version but Apple will only provide commercial HAP specs under NDA to a qualified business, not an individual hobbyist.

Re the definition of web server, fair enough -- I guess HomeSpan is a lightweight server.  However, I originally created my own parsers because of the somewhat, though not completely,  unique aspect of HomeSpan that requires connections to be permanently open.  For example, I looked into using Async Web Server, but its whole pitch is that is handles multiple incoming requests, processes them simultaneously, and closes the connections as soon as completed as fast as possible.  This is typical of most web servers, but is the opposite of what HomeKit wants.  It needs perpetually open connections, and it makes no sense to handle queries in parallel since the queries are not designed only to retrieve information but also to perform actions.  Some actions, such as locking a door, have built-in time constraints.  If HomeSpan received a conflicting query to open a lock while still processing a query to close the lock, the results would be strange.  This is not really an issue in practice, since HomeKit doesn't really operate like that anyway.  But I only raise it to show that the HTTP needs for HomeKit are somewhat different then a typical web server (though not necessarily a web SERVICE).

As for going forward, for a while now I've been trying to compartmentalize HomeSpan.  It's getting too big, both in footprint and in complexity.  And most people only use a small portion of the code for any given device.  For example, why have SpanPoint if you are not using it at all.   Folks who want to add MQTT, Bluetooth, etc., find they are often do not have enough memory to include everything they need.  

Ideally, I'd like to break HomeSpan into a central core into which components are added as standalone libraries from within the sketch when needed.  This is a little tricky to accomplish in the Arduino environment since the IDE compiles all C++ files within the library structure, whether or not you use them.  

The solution to this is to simply have separate libraries in different directories, but this would make things a lot more complicated for end users who use the IDE interface to pull down HomeSpan from the Arduino library manager.  But I could instead bundle things into a zip file that uncompressed into a multi-library structure.  This would work for anyone who pulls the code from GitHub directly.

One of the reasons I want to go down this path is so that other users can contribute directly to HomeSpan by adding HomeSpan-compatible libraries, rather than modifying HomeSpan itself.  This may require creating a small hook within HomeSpan every now and then, but once done each library could be separately supported and updated on whatever schedule needed, apart from HomeSpan updates.

I think your project would be a great candidate to try out this approach.  If I build the hooks mentioned above, you should be able to encapsulate your web server in a separate library.  In any given sketch, if the user wanted to use HTTPS, they would include both the HomeSpan library and your library.  Your library would presumably  include some start-up commands allowing the user to choose the port if needed.   For ease of use, I could create a dedicated HomeSpan web page with links to all such libraries (including "internal" ones, like SpanPoint, that I would peel off from the main branch).

Food for thought.

frankonski:
> by definition and Apple licensing, HomeSpan can only be used by hobbyists for their own personal devices.
Well, HomeSpan has an MIT license, so anyone can do pretty much what they want with the source. But is Apple involved in some way with HomeSpan? Or is it that you built HomeSpan while HAP specifications were easily available for whoever had an Apple Developer account? And for any hobbyist who want to take their project to the next level, that means HomeSpan has to be ripped off and replaced with something else... :(

I started looking at the process of being HomeKit certified. This is massive, especially that it seems to involve adding an Apple chip to my board. This is serious business..

> unique aspect of HomeSpan that requires connections to be permanently open [...] _HomeKit_ needs perpetually open connections

You're saying that HomeKit controllers (iPad, AppleTV, HomePod, etc) wants permanently opens TCP connections with their accessories? wow... didn't know that. Looks like [esp_http_server](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/protocols/esp_http_server.html) does support HTTP KeepAlive.

> the IDE compiles all C++ files within the library structure, whether or not you use them.
Yeah, that is one reason I don't use the Arduino IDE, though I think Arduino CLI is alike and is one issue I'd like to tacke in my project as well. After all, that seems to only be a wrapper around `g++`.

> In any given sketch, if the user wanted to use HTTPS, they would include both the HomeSpan library and your library.

That was my idea behind adding a `HomeSpanTLSWebLog.h` and `HomeSpanTLSWebLog.cpp`, that would _normally_ only be built into the final binary **only if** `HomeSpanTLSWebLog.h` gets included in the sketch itself. I'll play around that idea for a while and will push an updated draft PR if you want to look at the approach.

HomeSpan:
Re the certification, you don't actually need to Apple chip anymore.  It used to be that HAP encoding was done in Apple-provided hardware so without the chip it was impossible to connect to HomeKit.  However, this prevented a lot of manufacturers from making their devices HomeKit compatible - they were prepared to do something in software, but adding a proprietary apple-coprocessor to a TV, door lock, etc. requires significant manufacturing re-tooling.   Then in 2019 or so, Apple decided they would release the specs for HAP encryption and allow manufacturers to choose which method to use software encoding or Apple co-processor).

As an offshoot of this change, Apple was able to release HAP for hobbyists - since they had created a software-based method of encryption, they could share with hobbyists for their own use.  But they tweaked the non-commercial version so it is sufficiently different from the commercial version and the Home App can distinguish the two.   Apple will not provide the commercial version of the HAP API to anyone not engaged in a commercial capacity, even though the API would include lots of things that would be helpful for hobbyists, such as specs for the TV Service and any other new Services they add.

Re connectivity, the HAP protocol is basically one-way.  It is not possible for a HomeKit device to make a connection with the Home App.  The Home App must initiate the connection and keeps it open to enable the device to communicate event messages, such as "the user pressed a button to turn on the light, so please reflect this in the Home App".

Up until about the end of iOS 14, if you had 3 iPhones and 2 Macs, each would establish, and leave open, their own connection when you opened the Home App on each.  A lot of the logic in HomeSpan is keeping track of each connection, and knowing which connection has "subscribed" to listening for events the device needs to communicate.

It's such a low-level protocol that if you turned on a light using the Home App on one iPhone, it was the responsibility of the HomeSpan device to communicate this change of state to the other iPhones and Macs, assuming they Home App was open and a connection was established.

Each connection has their own separate encryptions keys, which are established whenever the connection is made/re-made.  You can see this in the L2-level log files when a connection is first established (it's called "Pair Verify").

Things got a lot simpler in iOS 15 when Apple launched their new HomeKit Architecture in which they mandated the use a Home Hub.  Now, regardless of how many iPhone, Macs, etc. you use, all HAP requests generally come from the central Home Hub.  But Apple does not guarantee that there are other connections they may sometimes need to be make, so HomeSpan still need to support the ability to retain more than one open connection.  But I rarely see more than two required.

Would be nice if Apple changed things to guarantee only one connection is needed so I can simplify a lot of the code, but it's been a few years and they have not made any further changes (they are mostly focused on Matter).

frankonski:
Thanks for the historical background about the chip, I wasn't aware of this.

> Apple will not provide the commercial version of the HAP API to anyone not engaged in a commercial capacity

_commercial capacity_. That is a chicken-and-egg problem. Assuming I want to build a new product and sell it. I don't have commercial capacity so I can't get the accessory to show up as "Certified". It looks clunky to try to sell a product that shows _Not Certified_. But to get the accessory certified, I need commercial capacity. The only solution is to sell my idea & concept to a company with such capacity.

> The Home App must initiate the connection and keeps it open to enable the device to communicate event messages, such as "the user pressed a button to turn on the light, so please reflect this in the Home App".

Ah, that I get. 

> they mandated the use a Home Hub

So that means Home App _may_ keep TCP connections open, but we shouldn't care because it now only connects to the hub, not to HomeSpan itself anymore. So the question is _does the hub / controller keep TCP connections established to its accessories / to HomeSpan? Or does the hub do interval polling with quick short-lived HTTP connections?_ I thought I saw the latter when turning on L2 log, meaning TCP/HTTP KeepAlive may not be necessary.

> they are mostly focused on Matter

Does HomeSpan become obsolete once Matter takes over? With Matter 1.2 having been ratified by major players few months ago (after I started learning ESP32 stuff, Arduino boards and HomeSpan!), do you see Apple moving away from HAP & "Apple-Certified Accessories" to something more generic like Matter?

HomeSpan:
@frankonski, I've added a new method `homeSpan.getLog()` that takes as an argument a callback function and then sends all the weblog output to that callback is chunks of 1024 bytes max (it uses hapOut with the callback you specify as the output handle).  The callback should be of type `void callback(const char *buf)`, where *buf* will be set to NULL at end of output.

Here is sample code that simulates an HTTPS request when you type `@w` into the Serial Monitor.   The output you get is an exact duplicate of what you would normally see if the browser requested the web log when you have log-level set to 2 (verbose).   If this method works, I can always peel off the first two lines that contain the HTTP header so you can replace with appropriate HTTPS.  Can also add a homeSpan() command so you can suppress responses to the normal HTTP web log commands as an option.

I hope this gives you everything you need to get started with a standalone TLS module.  Let me know how it goes.

```C++
#include "HomeSpan.h"

#define MAX_LIGHTS  2

void setup() {
 
  Serial.begin(115200);

  homeSpan.setLogLevel(1);
  homeSpan.enableWebLog(50,"pool.ntp.org","UTC");

  homeSpan.begin(Category::Lighting,"HomeSpan Max");

   new SpanAccessory();
    new Service::AccessoryInformation();  
      new Characteristic::Identify();

  for(int i=0;i<MAX_LIGHTS;i++){
    new SpanAccessory();
      new Service::AccessoryInformation();
        new Characteristic::Identify();
        char c[30];
        sprintf(c,"Light-%d",i);
        new Characteristic::Name(c);
      new Service::LightBulb();
        new Characteristic::On(0,false);
     WEBLOG("Configuring %s\n",c);
  }

  new SpanUserCommand('w', " - get web log test",webLogTest);     // simulate getting an HTTPS request for weblog

}

//////////////////////////////////////

void loop(){
 
  homeSpan.poll();
  
}

//////////////////////////////////////

void webLogTest(const char *dummy){
  Serial.printf("\n*** In Web Log Test.  Starting Custom Web Log Handler\n");     // here is where you would perform any HTTPS initializations   
  homeSpan.getWebLog(webLogHandler);      // this starts the normal weblog with output redirected to the specified handler (below)
}

void webLogHandler(const char *buf){
  if(buf!=NULL)
    Serial.printf("%s",buf);            // here is where you would transmit data to the HTTPS connection
  else
    Serial.print("*** DONE!\n\n");      // here is where you would close the HTTPS connection
}
```

frankonski:
Thanks for trying to help. Without having tried it yet, quick code review shows that weblogs over HTTP (non-TLS) would still be available, which is what I am trying to avoid.

As soon as I receive my new purchase (new ESP32 on its way) I will be able to give this a try.

HomeSpan:
Yes, as I noted in my previous comment I would add a command to suppress that.   Wanted to minimize changes to only what you need for POC.

frankonski:
I am still alive. Issues with new board now resolved, code compiles with different PINs using `#ifdef`, also using PlatformIO.
I now took some time to better understand your proposal, but that won't work because when I need access to the HTML produced by `getStatusURL`, the TCP/HTTP connection is already established.

I have an idea that will hopefully be less intrusive to HomeSpan source. And instead of being a PR to "Add TLS for weblogs", this will now be limited to "Allow status URL to send output to `httpd_req_t`" ..

frankonski:
> is to return the HTML that is generated from the Web Log.
Correct.
> You can use the new homeSpan.getWebLog(() method.
Not as it is, but I see where you are going with this. I can't use it now because the callback you added for my need does not have access to the `httpd_req_t *` pointer object that I have to use to write back the HTTP response. And I'm already in someone else's callback (I am in a `httpd_uri::handler` callback). However, if your callback declaration allowed us to provide user-specified data, and just pass this data back when calling the callback, that would work.

See below, which is a mix of real code (that I can't use), intertwined with suggestions in comments. If my suggestion is not clear I can create a new sample and a different pull request (or update this draft PR).
```
/*
  Callback that HomeSpan will call in chunks of 1024 byte. The callback signature
  is not helping at the moment.

  Suggestion of new callback declaration (see full example below):
  void homespan_weblogs_handler(const char *buf, void *user_data){...}
*/
void homespan_weblogs_handler(const char *buf){
  if(buf!=NULL) {
    // I don't want to send to serial, but good for debugging for now..
    Serial.printf("%s",buf); 
    
    // I need to send 'buf' to 'req', which is not available in this callback
    // Suggestion of full ballback signature and content:
    /*
      void homespan_weblogs_handler(const char *buf, void *user_data) {
        httpd_req_t *req = (httpd_req_t *) user_data;
        httpd_resp_send(req, buf, HTTPD_RESP_USE_STRLEN);
      }

    */

  }
}

/* HTTP handler for the "/logs/?" URI */
static esp_err_t weblogs_uri_handler(httpd_req_t *req)
{
  httpd_resp_set_type(req, "text/html");
  httpd_resp_send(req, "<h1>HomeSpan Weblogs</h1>", HTTPD_RESP_USE_STRLEN);

  /*
      This is where I should keep calling httpd_resp_send() or httpd_resp_send_chunk().
      However, calling this new homeSpan.getWebLog(callback) does not help.

      Suggestion: let users pass their own data, whatever that is. 'req' for me.
      homeSpan.getWebLog(homespan_weblogs_handler, (void *)req);
  */
  homeSpan.getWebLog(homespan_weblogs_handler);
  
  return ESP_OK; // This ends and closes the HTTP connection/
}

static const httpd_uri_t weblog_uri = {
    .uri       = "/logs/?", // kinda regex, will be handled by esp32's built-in httpd_uri_match_wildcard() method
    .method    = HTTP_GET,
    .handler   = weblogs_uri_handler
};
```

frankonski:
While trying to implement my suggestion I realize this does not compile... working on it.

HomeSpan:
Adding a second argument to pass an arbitrary structure should be easy to do - let me know when you are ready for me to add it.

frankonski:
Done. Can you check what is in my fork? https://github.com/HomeSpan/HomeSpan/compare/release-1.8.1-dev...frankonski:HomeSpan:getWeblogs_callback?expand=1

I can _almost successfully_ achieve my goal using only these few lines:
```
/*
  Callback that HomeSpan will call in chunks of 1024 byte.
*/
void homespan_weblogs_handler(const char *buf, void *user_data){
  httpd_req_t *req = (httpd_req_t *) user_data;

  if(buf!=NULL) {
    httpd_resp_send_chunk(req, buf, HTTPD_RESP_USE_STRLEN);
  } else {
    httpd_resp_send_chunk(req, 0, 0);
  }
}

/* HTTP handler for the "/logs/?" URI */
static esp_err_t weblogs_uri_handler(httpd_req_t *req)
{
  httpd_resp_set_type(req, "text/html; charset=utf-8");
  homeSpan.getWebLog(homespan_weblogs_handler, (void *)req);
  
  return ESP_OK; // This ends and closes the HTTP connection
}
```

The only hiccup is that when using this method of spinning up my own webserver (TLS or not) using `httpd_handle_t`, the library takes care of sending both the HTTP response code as well as the _ContentType_, which will be `text/html` by default. In my case above I have to explicitly set it to `text/html; charset=utf-8`. But whether I call `httpd_resp_set_type()` or not, the fact that `HAP.cpp:1124` resends the HTTP response code and content type, this has the effect of getting `HTTP/1.1 200 OK Content-type: text/html; charset=utf-8` being printed as text in the browser.

Not sure if HomeSpan should imply "if (callback) don't send HTTP response & content type" ...

HomeSpan:
Reviewing now...

HomeSpan:
I've reviewed the changes think they should work well.  Do you want to create a pull request and I'll merge?

Also, feel free to add the following just before HAP.cpp 1124:

```C++
if(callBack!=NULL)
```

to suppress the output of the HTTP header when using a callback.

HomeSpan:
It may be easiest to create a new PR for the changes --- we can then close out this PR and continue discussion in the new PR to keep things all together.

frankonski:
sounds good. New PR to come.

frankonski:
You recon this one could be closed? I can provide a sample on how to use the new getWebLog() and setup a separate web server, and setup TLS. Out of HomeSpan's direct functionalities.

HomeSpan:
Agreed - I'll close this one and will add some documentation about getWebLog() on the Logging.md page.  Would be terrific if you can post a repo of a simple example that shows how TLS can be added to a sketch.  For simplicity I recommend just using a simple single light-bulb (non-functional) as the Accessory.

Note I created 1.9.0 as the latest branch.  This should be functionally complete and I'll test for the rest of the week.  Absent any bugs I will deploy next weekend (and the 1.8.1-dev branch will be deleted).

PR #733, opened 2024-01-06
Fix crash in stream

mgeramb:
This fixes my crashes. For any reason using PSRAM in this class does not work.


HomeSpan:
This has been addressed.

PR #759, opened 2024-01-31
Adding defines to provide ability to use less flash memory.

frankonski:
I had to do that to move forward with my project. Looks like HomeSpan is taking *a lot* of flash, so adding the ability to save **8.3 Kb** of flash memory by building with one or all of these defines, disabling un-needed functionality.
```
-DHS_NO_INFO
-DHS_NO_HELP
-DHS_NO_CLONING
-DHS_NO_PAIRING_DATA
-DHS_NO_MEM_INFO
-DHS_NO_DETAILED_STATUS
```
**BEFORE**, the `processSerialCommand()` method can take up to 10.5 KB of flash.
flash_usage_before

**AFTER**, the same method uses 2.2 KB
flash_usage_after

If you like this idea / approach, I will add some help to the documentation to explain what HomeSpan functionaloty these defines disables.

### DIFF
--- a/src/HomeSpan.cpp
+++ b/src/HomeSpan.cpp
@@ -583,6 +583,9 @@ void Span::processSerialCommand(const char *c){
       LOG0("IP Address:        %s\n",WiFi.localIP().toString().c_str());
        if(webLog.isEnabled && hostName!=NULL)   
         LOG0("Web Logging:       http://%s.local:%d%s\n",hostName,tcpPortNum,webLog.statusURL.c_str()+4);
+#ifdef HS_NO_DETAILED_STATUS
+      LOG0("\nFor extra details, rebuild without -DHS_NO_DETAILED_STATUS\n");
+#else     
       LOG0("\nAccessory ID:      ");
       HAPClient::charPrintRow(HAPClient::accessory.ID,17);
       LOG0("                               LTPK: ");
@@ -614,6 +617,7 @@ void Span::processSerialCommand(const char *c){
       }
 
       LOG0("\n*** End Status ***\n\n");
+#endif
     } 
     break;
 
@@ -822,6 +826,9 @@ void Span::processSerialCommand(const char *c){
     break;
 
     case 'm': {
+#ifdef HS_NO_MEM_INFO
+      LOG0("HomeSpan memory info disable. Rebuild without -DHS_NO_MEM_INFO\n");
+#else
       multi_heap_info_t heapAll;
       multi_heap_info_t heapInternal;
       multi_heap_info_t heapPSRAM;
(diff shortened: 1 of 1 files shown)


HomeSpan:
A few comments on this:

* HomeSpan does take up a good amount of flash, but about 2/3 are from the ESP32 core itself.  For instance, compiling the ESP32 example WiFi Web Server with MDNS results in a flash size of 776,657 bytes.  The latest 1.8.1-dev branch compiles to 1,194,049 bytes, which means HomeSpan code itself takes about about 408 KB.
* The easiest way to expand flash is to use a different partition table.  The default table used in the Arduino IDE reserves a LOT of space for SPIFF storage, which is not used at all by HomeSpan.  If you do not need SPIFF storage for your project you can use the minimal SPIFF partition table and increase available flash from 1.3MB to 1.9MB - a huge jump
* The method of adding #defines is not scalable and does not properly work with the Arduino IDE, so I can't add into HomeSpan.  The reason is that in C/C++, #defines are ONLY applicable within the file in which they are defined, unless you place them in a header file and then include that header file multiple files.  BUT, this only works if you are compiling the entire source code and not using libraries.  If you are using a library, such as HomeSpan, you generally should NOT be compiling it every time.  This is why the ESP32 core library has lots of parameters that can't be changed unless you recompile the core, even though there are #defines all over the ESP32 core.   The reason why this works in PIO and other IDEs is that you must have things configured to recompile the entire HomeSpan library every time as if it was part of your sketch.  There is nothing wrong with this, but it will only work if you are compiling HomeSpan as part of your sketch, and not as if it were an external, self-contained library.

The ESP32 core within the Arduino IDE has a method of trying to pass pre-processor directives to the code when compiling by including those directives in a separate "build_opts.h" file placed in the same folder as your sketch.  I've explored this extensively but find it is not sufficiently robust since the #defines in that file will only be used if you somehow manually force the IDE to recompile the library.  This can be done, for example, by switching the versions of the ESP32 chip from S2 to C3.

Ironically, the idea of selective compiling and linking to reduce space is literally 50 years old and was solved 50 years ago in unix compilers, which separate the pre-processing, compiling, and linking phases of creating a program.   This generally gets encapsulated into "make" files, which are ubiquitous in unix.  But though many IDEs can use user-provided make files (except for the Arduino IDE), most people opt for allowing the IDE to figure out what to do with a multi-file program.  And the IDE often just slams everything together into a single file for processing, ignoring the whole concept of separate compilation and linking stages.

My plan is to try to broadly address the issue of flash size in the release after 1.8.1-dev (to be renamed 1.9.0 when published).  My goal is to allows users to include/exclude SpanButton, Pixel, OTA updates, etc. as needed for their sketch in a way that works across different IDEs.  I'm not sure this will be possible, but will explore.

HomeSpan:
Glad to help 😄

If you want to use OTA you indeed need to have a partition table that contains at least two OTA partitions.  HomeSpan checks for this and will warn you if you enable OTA without an appropriate partition scheme.  SPIFFs is mainly used to store pre-built web pages, graphics, or maybe to create an internal file system that works with standard C/C++ file functions.   You could probably eliminate that partition completely, though care needs to be taken that partition sections start on 4K boundaries.

PR #770, opened 2024-02-08
Allow user-defined data when receiving weblogs from callback

frankonski:
When using `homeSpan.getWebLog()`, this allows the caller application to pas arbitrary data that will become available to use in the callback which will receive the _string buffered in chunks_.

### DIFF
--- a/src/HAP.cpp
+++ b/src/HAP.cpp
@@ -256,7 +256,7 @@ void HAPClient::processRequest(){
     }
 
     if(homeSpan.webLog.isEnabled && !strncmp(body,homeSpan.webLog.statusURL.c_str(),homeSpan.webLog.statusURL.length())){       // GET STATUS - AN OPTIONAL, NON-HAP-R2 FEATURE
-      getStatusURL(this,NULL);
+      getStatusURL(this,NULL,NULL);
       return;
     }    
 
@@ -1095,7 +1095,7 @@ int HAPClient::putPrepareURL(char *json){
 
 //////////////////////////////////////
 
-void HAPClient::getStatusURL(HAPClient *hapClient, void (*callBack)(const char *)){                 
+void HAPClient::getStatusURL(HAPClient *hapClient, void (*callBack)(const char *, void *), void *user_data){
   
   char clocktime[33];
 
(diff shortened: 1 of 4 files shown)


HomeSpan:
I think this looks good.

frankonski:
I have changed this draft PR to a PR. But so you know, the way things work is a bit strange. While I setup a separate web server, be it HTTP or HTTPS using `http_handle_t`, the logic is so that one won't call `homeSpan.getWebLog()` until there is an actual connection to that separate webserver. And that means the standard weblogs served by `getStatusURL()` remains available, meaning the logs I want to protect over TLS remains available in clear text for as long as there are no connections done to the TLS server.

If I don't call `homeSpan.enableWebLog()`, none of the `WEBLOG` macro gets buffered.
If they were (which I kinda hoped when I disabled `enableWebLog()`), I would still not have NTP setup, which I need for my purpose.

The way I would approach this would be to add these APIs providing individual features that `enableWebLogs()` add:
- `homespan.maxWeblogEntries(int max)`: when not called, weblogs are not buffered, same as if a sketch does not call `enableWebLog()`. When called, weblogs gets buffered just like they are when `enableWebLog()` is called.
- `homespan.setNTP(const char *server, const char *tz)`: sets the NTP to use. Some sketch may require NTP without needing weblogs.

Then `homespan.enableWebLogs()` can be modified to call these two new methods. Then in my case I would end up calling the first two, and no longer call `enableWebLogs()`.

HomeSpan:
Don't worry about disabling the web log from being accessed from the normal HTTP server.  My plan is to disable that by having have the user enable webLogs as usual, but put a NULL into the last argument.  I'll then check for a NULL in HAP.c and stop the webLog call completely.  Will also indicate as much in the initial output on the Serial Monitor.

HomeSpan:
I was going to add the above anyway since there are some use cases where it may be helpful to use the NTP feature of weblogs simply to acquire the time, but not to enable actual weblogging.

frankonski:
> put a NULL into the last argument

good idea!

HomeSpan:
I just added functionality to allow a NULL value as the statusURL.  When specified as NULL, webLog.enabled remains FALSE, which means HTTP requests are not honored, and the Serial Monitor will NOT show that Web Logging has been enabled.

However, the timeServer and timeZone specified in the homeSpan.enableWebLog() call as the second and third parameters will still be called and used within the homeSpan.getWebLog() function.

Please give it a try and let me know if all is working.  Once you've verified I'll merge the PR.

frankonski:
Any issues merging this PR? I could test it with the NULL param.

HomeSpan:
All set.

frankonski:
So you know, it all works fine. Hitting the weblog URLs gives a 404. All good!

Issue #822, opened 2024-03-28
TempBuffer memory allocation failure should not use Serial.printf()


frankonski:
The new `TempBuffer` class will use `Serial.printf()` if memory allocation fails, then get into an infinite loop to stall the program.

The method `Serial.printf()` actually requires a mix of stack and heap memory, so there is a chance that the `Serial.printf()` text logged by HomeSpan may never get logged.

Back-to-back `Serial.print()` should be used instead.

This may be the reason why I posted the comment at https://github.com/HomeSpan/HomeSpan/issues/791#issuecomment-2025475526 about HomeSpan's CLI becoming fully unresponsive even though nothing got logged to the serial console.

HomeSpan:
Have you been able to confirm that this is indeed the problem.  If an alloc() fails, nothing bad happens.  The memory is simply not acquired.  There should still be plenty of heap for printf to produce the output.

Note that serial monitors are not fully stable.  If I leave mine up for a while and my MAC goes to sleep I often need to close and re-open the serial monitor for it to re-establish a USB connection.

frankonski:
> Have you been able to confirm that this is indeed the problem.

Not yet.

>  If an alloc() fails, nothing bad happens.

well, stalling the whole program in a `while(1)` loop with no logging (assuming this is the issue) is somewhat bad imo. The calling method failing to get a TempBuffer should decide how to handle this scenario, ideally.

> There should still be plenty of heap for printf to produce the output.

Ideally, I agree, yes.

> need to close and re-open the serial monitor for it to re-establish a USB connection.

That, I tried. No luck, no wake up, power still in D0. My only suspicion right now was these `while(1)`…

HomeSpan:
I'm not sure there is any better solution than at present.  Note that the esp32 log functions all use Serial.printf under the hood, so I do not think that is the issue.  Closing for now.

PR #823, opened 2024-03-28
Use standard print method on HomeSpan halt

frankonski:
Avoid using `printf()` during memory allocation failure, since `printf()` will also require heap.

This commit should also be cherry-picked in 1.9.0, for inclusion in a 1.9.1 release.

This is to address issue #822


frankonski:
Replaced additional `LOG()` with `Serial.print()` when HomeSpan halts (`while(1)`)

frankonski:
@HomeSpan , closed on purpose, no merge?

PR #826, opened 2024-03-31
Add Improv WiFi Serial Protocol Support

mccahan:
Long time listener, first time caller! Love the project, wondering if you'd like to include this thing I put together...

I haven't written any docs here since it's an unsolicited PR and I didn't want to do a lot of writing if you decided it wasn't a good fit for the project or want major changes; happy to write some docs if this is something you think is worth including.

**The Problem:**
I've been playing around with [rednblkx's HomeKey-ESP32](https://github.com/rednblkx/HomeKey-ESP32) project based on HomeSpan, which has a bit of an arduous setup process including installing and trying to troubleshoot a VSCode extension.

Lots of ESP projects like [WLED](https://install.wled.me/) use browser-based installers based on [ESP Web Tools](https://esphome.github.io/esp-web-tools/), which lets you flash and set up WiFi directly from the browser using web serial. I started a repo for handling building the binary for that project, but then I wanted it to be able to confirm the install and configure WiFi, which requires implementing the [Improv Wi-Fi standard](https://www.improv-wifi.com/) to contribute to either HomeSpan itself or to that project. I figured I'd try my luck here to see if you're interested.

**The Solution:**
I have a pass here at implementing Improv Wi-Fi serial support (sorry, my C++ is... rusty at best, it was never that great in the first place). Improv works by sending serial commands that start with "IMPROV1", so I tapped into the main loop serial command reader to detect those and pass them off to the Improv.cpp file for processing.

These classes handle:
- Responding to requests for device status (important when it comes to confirming install, or when you load up an ESP Web Tools instance)
- Returning the list of scanned WiFi networks to show in the ESP Web Tools UI
- Processing WiFi credentials and connecting

**The End Product:**
Once you have an image with the Improv stuff installed, you can go to any ESP Web Tools page (e.g. [their official demo](https://esphome.github.io/esp-web-tools/)) and configure/change WiFi.

I made a proof-of-concept [HomeSpan ESP Web Tools installer demo](https://mccahan.github.io/homespan-esp-webtools-demo/) (repo [here](https://github.com/mccahan/homespan-esp-webtools-demo)) that you can load up in a browser, flash a binary made using GitHub Actions, and configure WiFi straight from the browser.

That demo currently runs the 05-WorkingLED example, compiled in [GitHub Actions](https://github.com/mccahan/homespan-esp-webtools-demo/blob/main/.github/workflows/deploy-pages.yml) and pushed to GitHub pages. It compiles it, merges it the way web tools requires into a single binary, and serves it up for install directly from the browser.

WebInstallDemo

**Thing I Got Stuck On:**
Logging from within the improv class - LOG2 (etc) aren't defined on that scope, and apparently I can't just `::LOG2()` them based on how they're implemented? I can investigate more to get that debug logging in, right now it's just commented out.

**Future Idea:**
When you finish installing and setting up WiFi, ESP Web Tools allows you to return a URL, which it shows as a button on the ESP Web Tools screen. I was looking at setting this to link to a QR code generated so the user can just scan it to add to HomeKit, but that required touching some more stuff than I wanted to with this PR. But I do think that would make for a pretty slick setup process for projects.

### DIFF
--- a/src/HomeSpan.cpp
+++ b/src/HomeSpan.cpp
@@ -222,7 +222,7 @@ void Span::pollTask() {
   if(!serialInputDisabled && Serial.available()){
     readSerial(cBuf,64);
 
-    if(strncmp(cBuf, "IMPROV", 6) == 0) {
+    if(strncmp(cBuf, "IMPROV", 6) == 0) { // check for an Improv-Serial command, see https://www.improv-wifi.com/serial/
       processImprovCommand(cBuf, this);
     } else {
       processSerialCommand(cBuf);
--- a/src/HomeSpan.cpp
+++ b/src/HomeSpan.cpp
@@ -223,7 +223,7 @@ void Span::pollTask() {
     readSerial(cBuf,64);
 
     if(strncmp(cBuf, "IMPROV", 6) == 0) {
-      processImprovCommand(cBuf);
+      processImprovCommand(cBuf, this);
     } else {
       processSerialCommand(cBuf);
     }
--- a/src/HomeSpan.cpp
+++ b/src/HomeSpan.cpp
@@ -1159,6 +1159,12 @@ void Span::getWebLog(void (*f)(const char *, void *), void *user_data){
 
 ///////////////////////////////
 
+const char* Span::getDisplayName(){
+  return(this->displayName);
+}
+
+///////////////////////////////
+
 void Span::resetStatus(){
   if(strlen(network.wifiData.ssid)==0)
     STATUS_UPDATE(start(LED_WIFI_NEEDED),HS_WIFI_NEEDED)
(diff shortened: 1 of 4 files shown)


HomeSpan:
Thanks for this PR.  Though the functionality looks quite interesting, a web-based mechanism for loading sketches is outside the scope of what I intended for the project (at least at this time).  Your use of GitHub actions is really cool - I have never explored that aspect of GitHub.

HomeSpan:
Yes, thanks for the additional color.  However, I don't think adding this type of functionality is within the intended scope of HomeSpan at this juncture.

mccahan:
Got it, thank you. I'll see if there's another path forward on it externally.

I've seen so many cool HomeSpan-based projects out there gated behind "install Arduino IDE, download this archive, choose your board, flash it, use serial monitor or edit this config file or connect to the AP to set your WiFi credentials..." processes. The HomeKey one I found was very exciting and then I had to spend a few hours debugging PlatformIO's VSCode plugin not playing well with built-in Python on my machine, just extra friction points.

Appreciate all your thorough work on this project, just incredibly thorough documentation and all.

PR #858, opened 2024-06-02
Alert users to compatibility with Espressif Boards v3

JeremyLaurenson:
This checks the core_version definition of the espressif32 and throws an error with a description of the fix rather than failing on sodium include.

### DIFF
--- a/src/HAP.cpp
+++ b/src/HAP.cpp
@@ -26,6 +26,10 @@
  ********************************************************************************/
  
 #include <ESPmDNS.h>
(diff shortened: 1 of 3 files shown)


JeremyLaurenson:
Anyone who simply updates their IDE software board definitions or does a fresh install as I did will go chasing the issue (which I would assume soon will be 'everyone')
Generating a descriptive warning so that we dont all have to go chasing the solution will drive them to resolution far quicker.

PR #874, opened 2024-06-29
Add a `pause` function for servo

cccat6:
In my application, I wish the servo signal would pause when I am not actuating it. The new function implements my needs.


HomeSpan:
I like this idea and had not realized you can free-rotate an (analog) servo by shutting off the pulse train.  However, I'm not sure about the implementation of adding a new pause method with a boolean state, since the position is not controllable upon restoration.

Instead, I added logic to the existing ServoPin::set() method so it accepts a C++ NAN.  This sets the duty cycle to zero, allowing for free rotation.   To restore, just call set() again with an actual position as such:

```C++
ServoPin myServo(21);   // create servo on pin=21
myServo.set(20);    // set position to 20 degrees
delay(5000);   // hold position for 5 seconds
myServo.set(NAN);  // allow free-rotation
delay(10000);  // delay 10 seconds
myServo.set(20);  // reset position to 20 degrees
while(1);
```

Code change is implemented in the 1.9.1 branch and will be part of next release.  Please feel free to test and let me know what you think.

cccat6:
Thank you for taking the time to review my pull request and providing an elegant solution. I appreciate your implementation, which is indeed more concise and efficient than my initial proposal.

Regarding the concern about how users might want to pause the servo, I agree that using a duty cycle of 0 is a reasonable approach. While it's possible to consider using -Inf and +Inf to represent 0 and 1 duty cycles, I believe that a single pause state with a 0 duty cycle should suffice for most use cases.

Allow me to provide some context about my work to better explain the motivation behind my pull request. My primary goal is to optimize energy consumption for the entire system. To achieve this, I employed a transistor and MOSFET to control the servo's power supply. However, I encountered an issue where abruptly restoring power to the servo while the control signal was still being sent resulted in a high voltage peak, potentially damaging the servo. This problem arises because the motor inside the servo is an inductor, and the servo's digital logic shares the same power supply (VM and VCC). Unfortunately, this is an inherent limitation of the servo's internal design, and there is little I can do to address it through external circuitry. I discovered this issue after losing five servos.

The solution I found is relatively simple: introduce a short latency between powering on the servo and providing the control signal. This allows the servo power to stabilize before receiving commands.

Additionally, I wanted to enable manual control of the device while it is powered off. In scenarios where power consumption is not a primary concern, using a 0 duty cycle proves to be a suitable solution.

Thank you once again for your valuable feedback and the improved implementation. I appreciate your efforts in maintaining and enhancing the HomeSpan library. The library is really useful and I use it to build smart devices to make my life much easier.

HomeSpan:
Really glad you are finding the library useful.  Given the above, can I close this PR?

cccat6:
Sure, please.

PR #812, opened 2024-03-20
Adding support for warm and cold white pixel.

frankonski:
Tested with WS2811 IC chip, with a pixel led strip like the one at https://www.amazon.ca/BTF-LIGHTING-Addressable-Flexible-3000K-6000K-Decoration/dp/B0BNDRF325

Addressing issue #810

PR #897, opened 2024-07-28
Adding support for warm and cold white pixel.

frankonski:
Tested with WS2811 IC chip and a BTF-Lighting 5M DC24V Tunable 3000K-6000K led strip.
To replace older PR #812.

### DIFF
--- a/src/src/extras/Pixel.h
+++ b/src/src/extras/Pixel.h
@@ -53,6 +53,8 @@ namespace PixelType {
   pixelType_t BGRW={15,23,31,7}; 
   pixelType_t GBRW={15,31,23,7}; 
   pixelType_t GRBW={23,31,15,7};
+  // White-only pixels
+  pixelType_t WHITE_WxC={15,0,23,0}; // GBR, R=warm white, G=nothing, B=cold white
 };
 
 ////////////////////////////////////////////
@@ -90,6 +92,13 @@ class Pixel : public Blinkable {
         this->white=w*2.555;      
         return(*this);
       }      
+
+      Color WarmColdWhite(uint8_t warm, uint8_t cold)                 // returns Color based on 0-255 values of warm and cold white leds.
+      {
+        this->red = warm;
+        this->blue = cold;
+        return(*this);        
+      }
 
       bool operator==(const Color& color){
         return(val==color.val);
@@ -168,6 +177,7 @@ class Pixel : public Blinkable {
     
     static Color RGB(uint8_t r, uint8_t g, uint8_t b, uint8_t w=0){return(Color().RGB(r,g,b,w));}  // an alternative method for returning an RGB Color
     static Color HSV(float h, float s, float v, double w=0){return(Color().HSV(h,s,v,w));}         // an alternative method for returning an HSV Color
+    static Color WarmColdWhite(uint8_t warm, uint8_t cold){return(Color().WarmColdWhite(warm,cold));}  // an alternative method for returning an all-white Color
               
     int getPin(){return(rf->getPin());}                                                     // returns pixel pin if valid, else returns -1
     boolean isRGBW(){return(bytesPerPixel==4);}                                             // returns true if RGBW LED, else false if RGB LED


frankonski:
Should I test the `refact2` branch with my cool/warm white strip and this proposed change?

HomeSpan:
I'll add this in directly in an upcoming iteration later today or tomorrow since things have changed quite a lot and I'm still testing all the changes in my integrated setups.

Also, my plan is to add logic to handle color temperature mapping more directly.  

In the interim, can you fully test the existing RGB logic? Want to make sure the new mapping from RGB to GRB, etc. works as expected.  

In the end these changes to accommodate IDF 5 were a real pain.  Espressif stripped out too many of the low-level API functions to allow me to continue using by own interrupt driver.  But in using their driver you give up some really important flexibility.  In the end I was able to combine their driver with some low-level calls to reproduce the flexibility needed to support both Pixel and RFControl.  The good news is that as a result of the re-write, you can now use RFControl and Pixel in the same sketch (previously they were incompatible when used together).

HomeSpan:
After doing some additional research I finally found that RGB+CCT strips actually exist:  https://www.btf-lighting.com/collections/ws2805-rgbcct-24v/products/ws2805-pixels-led-strip-light-rgbcct-5in1-60leds-m-addressable-dual-signal-3000k-6500k

This strip provides control of **FIVE** distinct LEDS: Red, Green, Blue, Warm White, and Cool White.

In my original Pixel library I hardcoded colors as a single 4-byte integer.  But in my new library I simply use an 4-slot array of single-byte integers.  To accommodate the above type of device I will expand to use a 5-slot array, where the last two slots can be used for warm white and cool white, allowing full 5-color control.   This expansion is easy in my new library.

For your WW/CW-only device, the mapping will use the LAST two slots, instead of the first two.  This will allow me to add dedicated functionality to create color temperature adjustments similar to how we now have dedicated functionality to convert HSV to RGB.  Stay tuned for an update over the next day or so that incorporates the new logic.  Should be pretty cool (no pun intended).

frankonski:
Using PlatformIO, I had to force it to use arduino-esp32 3.0.2 this way in platformio.ini:
```
platform_packages=
  framework-arduinoespressif32 @ https://github.com/espressif/arduino-esp32.git#3.0.2
  framework-arduinoespressif32-libs @ https://github.com/espressif/arduino-esp32/releases/download/3.0.2/esp32-arduino-libs-3.0.2.zip
```
And during build, I get this error:
```
lib/HomeSpan/src/Network.cpp:30:10: fatal error: DNSServer.h: No such file or directory
```
Have you added a dependency on some DNS server library?

HomeSpan:
I'm using 3.0.4 but I think it should work with 3.0.2 as well. I have not added any new dependencies that are not already in the base Arduino-ESP32 library.  Can you try 3.0.4 just in case?

frankonski:
No luck with 3.0.4. Looking at your `Network.cpp` file, I don't understand your usage of the `dnsServer` variable. What is the intent?
Also, the only method called is `dnsServer.processNextRequest();` which, if I follow the documentation to it, shows me this API documentation:
```
  /**
     * @brief stub, left for compatibility with an old version
     * does nothing actually
     *
     */
  void processNextRequest() {};
```

HomeSpan:
Well, it was needed when originally placed there 2 or so years ago -- may not be needed anymore.   See if you can eliminate.   The Espressif folks don't really do a good job with keeping up docs and deprecating changed to alert users about older functions.

HomeSpan:
Was just reviewing the code again - note that the dnsServer is still used even though the process command is just stub.  DNS is started with the start() command and that does all the work.  It may be that the logic for the process() command was moved to the background at some point in the history of 2.0.x (or maybe even (1.0.x). 

But I think the DNS functionality itself is still needed by Network to allow the captive portal to function.

frankonski:
I removed the `include` as well as the `dnsServer` variable and usage. It now compiles fine, but dang...
```
Linking .pio\build\keeyes-esp32dev\firmware.elf
Retrieving maximum program size .pio\build\keeyes-esp32dev\firmware.elf
Checking size .pio\build\keeyes-esp32dev\firmware.elf
Advanced Memory Usage is available via "PlatformIO Home > Project Inspect"
RAM:   [==        ]  18.3% (used 60088 bytes from 327680 bytes)
Flash: [===Error: The program size (1436705 bytes) is greater than maximum allowed (1310720 bytes)
==*** [checkprogsize] Explicit exit, status 1
=====]  109.6% (used 1436705 bytes from 1310720 bytes)
```

HomeSpan:
Yeah, the Arduino library grew significantly in 3.X and you need to use a larger partition structure to fit HomeSpan.  I'll note this in the docs before I release the new version. 

The problem is that the Arduino library has transformed from a board manager to an application. For example, the library now includes its own support for Pixels.  Why?  Pixels are not part of an ESP32 chip.  True, they are commonly used, but an operating system library should only include things for the operating system and core functions, not applications that can be added on by users as needed.   

My guess is this trend will continue indefinitely - at some point it may make sense to build a HomeSpan-specific variant of the Arduino library but that comes with another set of complications.

frankonski:
So I'm now running into something somewhat strange, though I don't know who's at fault. The OTA update (from old code using HS 1.9.1) using newly HS 1.10, using the upload capabilities of PIO, it just stalls:
```
Uploading .pio\build\keeyes-esp32dev\firmware.bin
14:30:09 [DEBUG]: Options: Namespace(esp_ip='192.168.80.113', host_ip='0.0.0.0', esp_port=3232, host_port=33360, auth='homespan-ota', image='.pio\\build\\keeyes-esp32dev\\firmware.bin', spiffs=False, debug=True, progress=True, timeout=10)
14:30:09 [INFO]: Starting on 0.0.0.0:33360
14:30:09 [INFO]: Upload size: 1444240
Sending invitation to 192.168.80.113 
Authenticating...OK
14:30:09 [INFO]: Waiting for device...
14:30:19 [ERROR]: No response from device
*** [upload] Error 1
==== [FAILED] Took 13.55 seconds ====

```
If I switch back to the old branches, rebuild and restart the OTA update, the upload succeeds.

Is OTA from arduino-esp32 2.0.x to 3.0.x supported at all?

HomeSpan:
That's a good question - I have used 3.X OTA on an existing 3.X device but have not yet tried to use 3.X OTA to update a 2.X device.  I'll test this out and let you know what I find.

frankonski:
`*** OTA Error[1]: Begin Failed` is what I see (after I re-uploaded a HS 1.9.1 sketch with `LOG0` changed to `WEBLOG` for all OTA actions. _Lazy coding from couch, don't want to change my setup and connect the board via USB._

Maybe we should take this new issue in a separate Issue?

Note, I also though it could be related to changing the partition layout, but that was not the case. I could re-upload a new sketch with HS 1.9.1 with a larger partition layout, so HS 1.10.0 would then fit. No change when trying to OTA update with 1.10.0 and arduino-esp32 3.0.4.

frankonski:
I'll have to resort to USB and turn on ESP logging. `Updater.cpp` will log the details to the serial.

HomeSpan:
I've just tested and cannot reproduce - everything works fine.  I started with a clean sketch running under HS 1.9.1 using Arduino-ESP32 2.0.17.   I did use a large partition to begin with, since my sketch was already very large.  I then verified I could upload using OTA to make sure everything works in the baseline.

Next I updated the board manager to 3.0.4 and switched to HS 1.10.0.  I then uploaded to again using OTA and it worked as expected.  I was able to verify that the partition was changed from app0 (where it started) to app1, also as expected.

HomeSpan:
I tried again on a second device and this time it failed in the same way you described above.  I need to figure out why it did not fail on my first test device, but it may be related to something that was stored on the flash that was required by 3.0.4 since the first device I tested was not first erased.  I'll perform the same test on a fully erased device to see if that is the issue.

HomeSpan:
I have identified the source of the issue - it is unrelated to 3.0.4 and HS 1.10.0 and I can reproduce the issue all within 2.0.17 and HS 1.9.1, though I can't yet fully diagnose.

The problem seems to be with sizing and partitioning.  If I upload a sketch that is bigger than the partition CURRENTLY on the device via OTA, I get a failure at the start of the OTA, even if I change the partition scheme to be larger.  This makes sense because OTA uses the existing partition on the device, not the new one you are uploading.

The solution to this SHOULD be that you need to first use OTA to upload the original sketch and also change the partition table.  But this does not work.  Instead I need to upload with Serial and change the partition table.  Only at that time does the partition table seem to be properly updated.

I can see why OTA might not allow you to change the partition table via OTA itself, but it also may be a bug in how OTA works.  In any case, the only workaround I have now is to use Serial to upload the sketch with a new partition table.  You can upload the original sketch under 2.0.17 or the new sketch under 3.0.4.  Does not matter and OTA under 3.0.4 will begin working in either case once the partition table is expanded.

frankonski:
> The solution to this SHOULD be that you need to first use OTA to upload the original sketch and also change the partition table. But this does not work.

Same here, I tried it yesterday. 

Also, looks like all esp_logs, even though fully turned on at verbose level at starts, gets disabled (by who?) right after wifi connects. I will see if I can add a wifi connect callback, re-set full esp verbose logging, and get something logged by espidf.

HomeSpan:
I found that we are not the only one with this issue.  The ESP forum discusses the root of the problem, which is that OTA does not support changes to the partition table.  Others have figured out some clever workarounds by rewriting the partition table during run-time, but apparently this can be risky.  I may look into seeing if that is possible.  For now, can you create a separate issue on this topic so we can finish up on the pixel stuff and hopefully close that out once tested.

frankonski:
ok. Meanwhile, here's the proof when monitoring a small partition HS on 2.0.17 IDF when doing an OTA to 3.0.4 with larger size:
```
[341821][D][Updater.cpp:139] begin(): OTA Partition: app1
[341826][E][Updater.cpp:163] begin(): too large 1444240 > 1310720
[341832][E][ArduinoOTA.cpp:252] _runUpdate(): Begin ERROR: Bad Size Given
```

frankonski:
Back to the issue at stake: warm/cold pixel. 
After a fresh USB upload, and using something similar to your example, I get somewhat crappy results:
1.  I can see alternating warmer and colder "pixel" (a pixel / IC chip is 48 LEDs, 24 warm and 24 cold).
2. Changing the temperature through the app (Eve or Home) gets randomly applied to the strip.

Do you have such strip handy?

With my draft PR, going from warm to cold does not show alternating pixels. I can attach picture of the results if you want, and share the entire code. That light strip code is not a _commercial secret_, but the only way I could find to illuminate my house in a custom way.

HomeSpan:
I have a full RGBWC strip with 5 LED colors per pixel.  I can seamlessly blend from cool-white to warm white using the new library.

Did you look at my new Pixel_RGBWC example?

HomeSpan:
Remember you will need to use setTiming if your pixels have a different timing, as well as provide the "order" in which the bytes will be sent.  For a strip with only two LEDs, this probably means the manufacturer used a 3-channel PWM chip with one of the wires not connected.  You'll need to set the pixel type to something like "CW-" or "WC-", and then use the either Pixel::WC() to set the warm and cool LEDs from 0-255, or use the new Pixel::CCT() to simply set the temperature, provided you first specified the temps using Pixel::setTemperatures().

To debug I recommend starting with just a simply script the sets the LEDs without using HomeKit - this makes it easier to test the library without needing to worry if the HomeKit conversion from temperature is correct.

frankonski:
Maybe this is it. You have a full RGBWC strip. I have a WC strip only.
I used the example at https://github.com/HomeSpan/HomeSpan/blob/release-1.10.0/examples/Other%20Examples/Pixel-RGBWC/Pixel-RGBWC.ino
When I tried to initialize the strip as "WC", it failed as `byteCount` was too small. I stopped there and tried to initialize the Pixel as "RGBWC".

Should I rather try to make it work by initializing as "WC"?

> You'll need to set the pixel type to something like "CW-" or "WC-"

Oh! I'll look further into this. Thanks!

HomeSpan:
Yes, sorry I did not yet have a chance to update the docs.  The way these strips seem to work is that the manufacturer either uses LEDs with embedded PWM chips, or they use separate chips to control one or more standalone LEDs.  I have found PWM controllers with 3, 4, and 5 channels, and even one with 12 channels.  For example, Adafruit sells a 3-channel device but instead of connecting each channel to Red, Green, or Blue they connect all three channels to a separate white LED.  They then direct everyone to use an RGB library and just realize that the lights are white no matter what.

In your case I'm guessing the manufacturer used a three-channel PWM controller and left one channel unconnected. This means the strip requires THREE bytes for each pixel control, even though one will be ignored.  In my new library this is specified by setting the type ad "WC-", "C-W", "-CW", etc.   The "-" is represents the placeholder to send a dummy byte (I send whatever it in the "R" color, but it does not matter.  For example, if you know the third channel is disconnected on your light strip, you can specify "WC-", "WCR", "WCB", or even "WCW".  The third byte has no effect, but it MUST be transmitted else the strip will drop one of the real colors you are trying to send (resulting in the alternating pattern you see).

frankonski:
This is what I had figured out based on my PR:
```
  // White-only pixels
  pixelType_t WHITE_WxC={15,0,23,0}; // GBR, R=warm white, G=nothing, B=cold white
```
So I ended up initializing the Pixel class with `"W-C"`. But now the temperature color selected in the app no longer matches the strip color. 

That was my formula. Commented lines are my working code.
```
      float warmVal = (px->T.getNewVal<float>()-140) / 360 * 255;
      float coldVal = 255 - warmVal;
      
      px->pixel->set(
//        Pixel::WarmColdWhite(static_cast<uint8_t>(warmVal / 100 * px->V.getNewVal<float>()),
//                             static_cast<uint8_t>(coldVal / 100 * px->V.getNewVal<float>())),
          Pixel::WC(static_cast<uint8_t>(warmVal / 100 * px->V.getNewVal<float>()),
                    static_cast<uint8_t>(coldVal / 100 * px->V.getNewVal<float>())),
        px->nPixels);
```
With color temperature range of
```
    T.setRange(140,500,1);
```

HomeSpan:
I'm not sure I'm following your code exactly.   Give the new CCT command a try based on the example.  If the color temp seems to be in reverse, it may mean your strip is actually C-W, not W-C.

frankonski:
Yeah, sorry, it took me a while to figure this one out.

Using HSV, full bright colors are represented on a 360 degree circle. 
- 0 and 360 are full bright red.
- 120 is full bright green.
- 240 is full bright blue.

In my test with HS 1.9, I found that if I used your GBRW pixel type (`pixelType_t GBRW={15,31,23,7};`) and set the second and fourth array values to 0 (`pixelType_t GBRW={15,0,23,0};`), then using the Home/Eve app, `Characteristic::ColorTemperature` would return a value I can use on this 360 degree circle.

On full red color in the app, I would get full warm white capacity.
On full blue color in the app, I would get full cold white capacity.

Assuming a scale between 0 and 100, if I want a smooth transition from warm to cold, and cold to warm, this is how I'd need to set the leds
* Full warm = 100% red, 0% blue.
* Full cold = 0% red, 100% blue.
* Half and half = 50% red, 50% blue
* A bit warmer = 75% red, 25% blue.
* ... end so on.

On this 360 degree circle, how do you calculate these percentages, on a scale between 0 and 255 (led values)? That's the formula given the value returned by `T`.

```
      float warmVal = (px->T.getNewVal<float>()-140) / 360 * 255;
      float coldVal = 255 - warmVal;
```
Then apply some brightness to the values calculated: decrease the 0-255 value calculated based on the percentage of brightness.

HomeSpan:
I see, though this is quite convoluted.  ColorTemp is a Mirek's, which is a value from 140-500.  To convert to temp you divide into one million.  This means Kelvin = 1,000,000 / ColorTemp.  I don't know why Apple choose to use the Mirek scale instead of Kelvin, but that's what they decided.

The change you made in 1.9.1 (pixelType_t GBRW={15,0,23,0};) is a workaround to get the Pixel class to skip a byte, but it's only by luck of how I wrote the original code that it works.  Technically the second zero causes access to an array that would be out-of-bounds and yield a core dump, but this is mitigated by something else by coincidence. 

Recommend you use CCT in the new library - it's designed to handle your exact situation.

frankonski:
Made some progress, but still no good. Initializing with `Pixel testPixel(PIXEL_PIN, "CW-");`, I can get my test scenario to _almost_ run with no hiccups:

```
#define MAX_BRIGHTNESS  255
/**
 * FCOB WS2811 IC CCT LED Strip, from BTF Lighting.
 * 10W per meter.
 * One WS2811 IC chip drives 48 LEDs (24 warm, 24 cold). 
 * 12 IC per meter, 576 LED per meter. Total of 2880 LED on 60 IC chips.
*/
#define NPIXELS   60*1 // 60 pixels x number of strip

void flashWhite(bool warm, bool cold) {
  for(int i=0;i<MAX_BRIGHTNESS;i++){
    testPixel.set(Pixel::WC(i*warm,i*cold),NPIXELS);
    delay(10);
  }
    
  for(int i=MAX_BRIGHTNESS;i>=0;i--){
    testPixel.set(Pixel::WC(i*warm,i*cold),NPIXELS);
    delay(10);
  }

}

void flashGradiant(){
  int c=0, w=0;
  bool out=false;
  while(out==false){
    if (w < MAX_BRIGHTNESS && c==0) {
      w++;
      testPixel.set(Pixel::WC(w,c),NPIXELS);
      delay(10);
    } else if (w == MAX_BRIGHTNESS && c < MAX_BRIGHTNESS) {
      c++;
      testPixel.set(Pixel::WC(w,c),NPIXELS);
      delay(10);
    } else if (c == MAX_BRIGHTNESS && w > 0) {
      w--;
      testPixel.set(Pixel::WC(w,c),NPIXELS);
      delay(10);
    } else {
      c--;
      testPixel.set(Pixel::WC(w,c),NPIXELS);
      delay(10);
      if (c<=0) out=true;
    }
  }
}

void loop(){
  Serial.printf("Warm...");
  flashWhite(1,0);

  Serial.printf("Cold...");
  flashWhite(0,1);

  Serial.printf("Dual...");
  flashWhite(1,1);

  Serial.printf("Gradiant...");
  flashGradiant();
} // end of loop()

```
The only issue is when it get to almost 100% brightness of _Dual_, the strip flickers.
And then, if I use "CW-" in HomeSpan, nothing really works. Brightness won't change (it would actually warm up or cool down the color), tapping on the preset colors changes to a completely different color, or does not respond.. That same code works just fine with HS 1.9 + my PR.

HomeSpan:
If it flashes at maximum brightness it means your power supply is not providing enough current.   I have the same thing and tested with different supplies to verify this is the problem.   I will send you some code to use with HomeSpan.  If pixel below work, then HomeSpan will as well..

HomeSpan:
Assuming "CW-" is correct as per your test above, the following code should implement colorTemp control in HomeSpan.  Note that below I set the number of pixels to 5 --- I believe you have more, but you should start with a lower number to avoid overloading your power supply.  You can then increase as desired.  The code below interpolates the temperature between the two LEDs, so both LEDs are never fully maxed-out at the same time.  This probably will prevent any power-supply issues even if you set NPIXELS to the full strip.

```C++

#define NEOPIN          23        // pin number to which NeoPixel strand is attached
#define NPIXELS         5         // number of controllable pixels in strand (which may be less than actual number of LEDs)
#define WARM_LED_TEMP   3000      // temperature (in Kelvin) of Warm-White LED
#define COOL_LED_TEMP   6500      // temperature (in Kelvin) of Cool-White LED
 
#include "HomeSpan.h"

///////////////////////////////

struct NeoPixel_CW : Service::LightBulb {      // define WC Lightbulb to control the Warm White and Cool White LEDs on the NeoPixel Light Strip
 
  Characteristic::On power{0,true};
  Characteristic::Brightness V{100,true};
  Characteristic::ColorTemperature T{(uint32_t)1.0e6/((COOL_LED_TEMP+WARM_LED_TEMP)/2),true};     // set initial value to average of Warm and Cool Temp
  
  Pixel pixel{NEOPIN,"CW-"};
  
  NeoPixel_CW() : Service::LightBulb(){

    pixel.setTemperatures(WARM_LED_TEMP,COOL_LED_TEMP);        // set color temperatures of Warm-White and Cool-White LEDs
    
    V.setRange(5,100,1);
    T.setRange(1.0e6/COOL_LED_TEMP,1.0e6/WARM_LED_TEMP);      // set range of control to match range of Warm-White and Cool-White LEDs
    
    update();                                 // manually call update() to set pixel with restored initial values
  }

  boolean update() override {

    Pixel::Color colorWC=pixel.CCT(1.0e6/T.getNewVal(), V.getNewVal()*power.getNewVal());   // convert HomeKit temperature (in Mirek) to Kelvin
    pixel.set(colorWC,NPIXELS);
    return(true);  
  }
};

///////////////////////////////

void setup() {
  
  Serial.begin(115200);
 
  homeSpan.begin(Category::Lighting,"WarmCool Light");

  SPAN_ACCESSORY("Neo CW");
    new NeoPixel_CW();
}

///////////////////////////////

void loop() {
  homeSpan.poll();
}

///////////////////////////////
```

frankonski:
ok, I think things are now working fine. I don't know what went wrong with my previous use of HS 1.10, but after double-checking all my code, things are now working. I must have had some incorrect init values or something like that. 
There is a tiny flickering on half-warm, half-cold, but I was also noticing it with my previous PR, which I am now leaning to think that is is caused by the length of my 22 AWG wire, as the power supply I have is a 24v 6amp.
.. or the size of the wires I soldered on my prototype board?? will have to check.

But I would say, over & all, for now, HS 1.10 seems to be doing what it should with WC LED strips!

oh, and mine shows 3000K-6000K on the pack, not 6500K, as an FYI.

And question: who decides of the initial default pre-set values for the temperature in the Home app?

HomeSpan:
Really glad to hear it's working.  The range of temps is set by the user using setTemperatures() as per the above code.  There are default values if not set, but that's only there as a convenience in case you don't know the temps for your lights.  Normally, you should always be specifying them using setTemperatures().

The code above always changes the range of the colorTemperature Characteristic from its Apple defaults of 140-500, to match those derived from setTemperatures().  This is not required but it ensures the full range of the color temperature graphic in the Home App and Eve app are used (in Kelvin's, the Apple default range of 140-500 maps to 2,000-7,142 K, which is often beyond the range of most lights).

Finally, the code also sets the initial colorTemperature the first time you boot to be the average of the two LED temperatures you specified, which should result in 50% warm and 50% cold.  But after the first clean boot the latest value is stored and retrieved in NVS and will always be set to whatever the last value you used.

frankonski:
Thanks. My question "who decides of the initial default pre-set values for the temperature in the Home app?" was more about the 3 pre-defined (labeled as _Favorites in Eve_) color buttons.

HomeSpan:
That logic is all within the Home App and seems to change from time to time.  At present I think the temps are saved as favorites every time you set a new temp color in the Home App.  Though I'm not sure the exact mechanism.

frankonski:
thanks.
oh, regarding
> Really glad to hear it's working.

That is with `#include <DNSServer.h>` and the class variable `DNSServer dnsServer;` removed from `Network.cpp`.

HomeSpan:
I don't think we can remove DNS - have you checked that the HomeSpan setup page works without the DNS?

frankonski:
In AP mode? I have to admit I have never tried that functionality before.. have to give it a try.

HomeSpan:
See the HomeSpan Users Guide for details on how it gets used.  You can start the AP via the Control Button, or by typing "A" into the CLI.  Lots of folks use this to set-up their device.  The method uses standard captive-portal stuff, which it archaic in many ways but no one has ever come up with anything better that is not proprietary (Espressif has their own method and app, of course).  I did not want anyone to have to use a separate App on their phone, so went the captive-portal route.

What's ridiculous is that HomeKit support native integration using a QR-Code. But only in the COMMERCIAL version.  Apple turned this feature off for the non-commercial version of HAP used by HomeSpan.

frankonski:
I bought Meross dimmer switches that didn’t require an app to be added to Home, and didn’t have to read a QR code. Dunno if the switch is using BLE to Home.app to transfer its name and accessory type, but the icon automatically appeared in the Home app when adding a new accessory, and I only had to type its 8-digit number. The switch then joined my wifi without me telling it which one to connect to (nor its password!). Another discussion for another thread. 
Will try the captive portal soon.

HomeSpan:
The QR Code is just another form of the Setup Code.  Same thing.  The key is that automatic provisioning of WiFi Credentials it built into the Home App. But just for commercial devices, which is really too bad.  Wish Apple would release all of Home Kit for non-commercial use.

frankonski:
ok, so after fiddling around with VSCode and PIO, I finally found a solution which hopefully can be implemented. The only way I was able to `DNSServer.h` found was to add `#include <DNSServer.h>`  in `HomeSpan.cpp`, right below `#include <ESPmDNS.h>`.

Maybe it can be placed somewhere else in `HomeSpan.cpp`, but having it only in `Network.cpp` makes PIO to never be able to find the header file.

And to answer your question: No, captive portal does not work without DNS server. I tried to find a way to redirect DNS queries using `MDNS`, but didn't succeed and stopped.

And I can understand why captive portal is used: it kicks open a browser as soon as you connect to the AP.

There is an issue with Captive Portal on Windows 11, but I will report that in a separate issue, even though I don't use captive portal functionalities (for now).

frankonski:
Just checking, will you be integrating the change above so things work out of the box with PlatformIO users?

HomeSpan:
Will be offline for a few weeks.  Will carefully review then - not planning to release 1.10 (or likely remained to 2.0) until then in any case.

frankonski:
Feel free to close this PR since support for WC pixel strips works "out of the box" with 2.0.0rc1 (except that header file issue with PIO which we can track separately).

Issue #899, opened 2024-08-04
hapOut sporadically gets into a fail state caused by WEBLOG thread safety


frankonski:
While trying to figure out why hitting weblogs was not always sending the full HTML page, and why sometimes the Home app returns "No response", I ended up adding a bit of debug in `HAPClient::getStatusURL()` as such:
```
  hapOut << "</body></html>\n";
  hapOut.flush();
  // Let's see if there was an error flushing the stream
  if(!hapOut.good()) {
    LOG0("Error flushing stream. (eofbit: %d, failbit: %d, badbit: %d)\n", hapOut.eof(), hapOut.fail(), hapOut.bad());
  }
```
And the outcome I get, from time to time, is
```
Error flushing stream. (eofbit: 0, failbit: 1, badbit: 1)
```
No idea why at the moment, and no TLS code, just standard weblogs.

And there does not seem to be a way to recover from this other than reboot the device.

Hitting Refresh in Chrome gives 
```
192.168.1.106 didn’t send any data.
ERR_EMPTY_RESPONSE
```
and turning L2 logs, I get this:
```
=======================================
** Client #2 Connected (773 sec): 192.168.1.116

<<<<<<<<< 192.168.1.116 <<<<<<<<<
GET /status HTTP/1.1
Host: 192.168.1.106
Connection: keep-alive
Pragma: no-cache
Cache-Control: no-cache
Upgrade-Insecure-Requests: 1
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/127.0.0.0 Safari/537.36
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.7
Accept-Encoding: gzip, deflate
Accept-Language: en-US,en;q=0.9,fr;q=0.8
------------ END BODY! ------------

>>>>>>>>>> 192.168.1.116 >>>>>>>>>>
Error flushing stream. (eofbit: 0, failbit: 1, badbit: 1)
------------ SENT! --------------
=======================================
** Client #2 Connected (773 sec): 192.168.1.116

** Client #2 DISCONNECTED (773 sec)
```

frankonski:
`hapOut` is global right? Wouldn't it need to be protected against concurrent access, like when a controller requests an update from HomeSpan, and when a user hits the weblog?

[...]

PR #901, opened 2024-08-07
Making WEBLOG thread safe

frankonski:
Related to #899, here is a draft PR that makes WEBLOG / vLog thread safe.

### DIFF
--- a/src/HAP.cpp
+++ b/src/HAP.cpp
@@ -1180,12 +1180,19 @@ void HAPClient::getStatusURL(HAPClient *hapClient, void (*callBack)(const char *
   hapOut << "<p></p>";
   
   if(homeSpan.webLog.maxEntries>0){
+    int currentEntries=0;
     hapOut << "<table class=tab2><tr><th>Entry</th><th>Up Time</th><th>Log Time</th><th>Client</th><th>Message</th></tr>\n";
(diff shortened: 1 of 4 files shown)


HomeSpan:
Thanks --- I will try include something like this in the next release.   But can't accept or spend any time on PRs at present since code base will be changing much too much.  Pixel class will not work under IDF 5 unless I use the legacy driver, and that can be deprecated at any time by Espressif.  The new RMT library from Espressif is not very flexible and way too abstract.   Rather than create an operating system library, they seemed to have created a much higher-level interface that masks too many of the functions of the chip.

I will be spending the next weeks to months refactoring the existing Pixel and RFControl libraries using the underlying native RMT functions (as opposed to the abstract driver library).  Won't be making any updates to software til then, and structure of rest of program can change without notice.

Under Arduino-ESP32 3.X the HomeSpan footprint grew considerably (without any changes) and HomeSpan no longer fits into a standard partition.  One of the reasons the code will be changing a lot is that I'm going to try to separate out a lot of optional functions so that the footprint can be smaller.  Among other things, this will also effect WebLog, since that will be one of the optional components.

frankonski:
So, no bugfix until refactor is complete? Same with # 880 and # 897?

HomeSpan:
Correct, no changes for now.

PR #983, opened 2025-01-09
Adding typedef to Characteristics enum values.

frankonski:
This allows using enums as return values, such as this:
```
  Characteristic::SecuritySystemCurrentState::SecuritySystemCurrentStateEnum_t getState() {
    return partitionCurrentState->getVal<Characteristic::SecuritySystemCurrentState::SecuritySystemCurrentStateEnum_t>();
  }
```
Else we have to rely on returning the integer value, losing the portability of using constant names.

### DIFF
--- a/src/Span.h
+++ b/src/Span.h
@@ -475,7 +475,7 @@ namespace Service {
 // Macro to define Span Characteristic structures based on name of HAP Characteristic, default value, and min/max value (not applicable for STRING or BOOL which default to min=0, max=1)
 
 #define CREATE_CHAR(TYPE,HAPCHAR,DEFVAL,MINVAL,MAXVAL,...) \
-  struct HAPCHAR : SpanCharacteristic { __VA_OPT__(enum{) __VA_ARGS__ __VA_OPT__(};) HAPCHAR(TYPE val=DEFVAL, boolean nvsStore=false) : SpanCharacteristic {&hapChars.HAPCHAR} { init<TYPE>(val,nvsStore,MINVAL,MAXVAL); } };
+  struct HAPCHAR : SpanCharacteristic { __VA_OPT__(enum Value_t {) __VA_ARGS__ __VA_OPT__(};) HAPCHAR(TYPE val=DEFVAL, boolean nvsStore=false) : SpanCharacteristic {&hapChars.HAPCHAR} { init<TYPE>(val,nvsStore,MINVAL,MAXVAL); } };
 
 namespace Characteristic {
   


frankonski:
@HomeSpan, should this one go in the dev branch instead, for now? I think this is the 4th time I've been trying to get this change in.

HomeSpan:
I see what you are trying to do, but I'm not sure about the implementation.  First, note that the enums are all integers, so even if you return an *int* from a function, you can still use the enum constants to compare to that *int* at any point.  The primary purpose of creating a named enum is for readability only, since the enum structure does not prohibit assignment of integers that don't match a specific enum.  To create a true type-safe framework we'd have to use *enum class*, which we can't because these Characteristics really are integers and HAP expects them to be interpreted and transmitted as integers.

For example, the following snippet compiles and works as expected to return the number 6, even though it is not defined within the enum:

```
enum Color { RED, GREEN, BLUE };

Color getColor(){
  return((Color)6);
}
```

Nevertheless, there are advantages to naming the enums for readability as mentioned.  However, there are two changes I'd recommend to your implementation.  First, since the enums are already scoped within each Characteristic, there is no reason to create a unique enum type for each Characteristic. 

So instead of:
```
    return partitionCurrentState->getVal<Characteristic::SecuritySystemCurrentState::SecuritySystemCurrentStateEnum_t>();
```

it would be easier to say something like:

```
    return partitionCurrentState->getVal<Characteristic::SecuritySystemCurrentState::Value_t>();
```

where *Value_t* (or just *Value*, or *Val*, or whatever) is commonly used for each Characteristic.  There is no danger of confusion with regard to which Value_t is being referenced since you are already providing the fully-qualified name, so in the example above you know it is for the SecuritySystemCurrentState Characteristic.

The second change I'd make is to simply name the enum, instead of using typedef.  Typedef enums (as well as typedef structs) are a holdover from C.  Though typedef is still used in C++, it's generally not used for things like enums and structs.  For example, these are basically the same (from the end-user perspective, not necessarily from the compiler and linker):

```
typedef enum { RED, GREEN, BLUE } Color;              // C style - creates an alias, Color,  to an unnamed enum
typedef enum Color {RED, GREEN, BLUE} Color;     //  C style - creates an alias, Color, to a named enum, Color
typedef enum Color {RED, GREEN, BLUE} Color_t;   //  C style - creates an alias, Color_t, to a named enum, Color
enum Color { RED, GREEN, BLUE };                            // C++ style
```

The C++ style enum is completely valid in C, but in C you would be required to declare a Color variable as such: `enum Color myColor` whereas in C++ you can simply say `Color myColor`.  If you use the typedef versions above in C, then you can simply say `Color myColor` just like in C++.

My recommendation is therefore to make a simple change by adding the name *Val* to the enum:

```
struct HAPCHAR : SpanCharacteristic { __VA_OPT__(enum Val {) __VA_ARGS__ __VA_OPT__(};)... 
```

Try that out and if it works well for you update the PR, point to the dev branch, and I'll add in next release.

frankonski:
I suck at C++, I'm a C guy..
While it compiles fine, I do not get expected behavior when using
```
  struct HAPCHAR : SpanCharacteristic { __VA_OPT__(enum Value_t {) __VA_ARGS__ __VA_OPT__(};) HAPCHAR(TYPE val=DEFVAL, boolean nvsStore=false) : SpanCharacteristic {&hapChars.HAPCHAR} { init<TYPE>(val,nvsStore,MINVAL,MAXVAL); } };
```
and
```
  Characteristic::SecuritySystemCurrentState::Value_t getState() {
    return partitionCurrentState->getVal<Characteristic::SecuritySystemCurrentState::Value_t>();
  }
```

`Val` in itself seems to be a reserved word, so I went with `Value_t`. Like I said, things compiled fine, but nothing was working. I will dig deeper when I can, but I am fine with your suggested approach (as long as it works!).

frankonski:
wait, hold on. It is with HS 2.0 that my stuff is not yet working. I have to get it working fine before trying your suggestion! Until further notice, I have to keep using my build using HS 1.9.1.
:(

frankonski:
PR updated. My stuff on the 1.9.1 branch works like a charm with your suggestion (my change in 1.9.1 is tracked at https://github.com/frankonski/HomeSpan/commit/3e946eb0d9d6cf65a037c856fda17b71e472ca4d)

HomeSpan:
PR added to dev branch.  Thanks!

PR #994, opened 2025-01-23
Add function useEthernet()

mgeramb:
In my app does the autodection of ethernet now work.
The reason is, I have to start the network to get the KNX configuration. Depending on that, I have to create HomeSpan or not. But for this reason, Ethernet is already started and the event for starting is already sent, and HomeSpan does not get the start event.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,9 @@ Requirements to run HomeSpan depend on which version you choose:
  
 * **Bug Fixes**
   * Fixes a latent bug that prevented compilation when the homeSpan methods `controllerListBegin()` and `controllerListEnd()` were used in a sketch
+ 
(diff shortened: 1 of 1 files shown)


HomeSpan:
This is a good solution - I will incorporate into next release.

HomeSpan:
Did you still need this function - I noticed the commit was pulled back.  Also, can you provide some more details on exactly how you are using HomeSpan?  In the dev branch the event handler is configured before the main Arduino `setup()` function is called, so as long as you are configuring your ethernet board inside `setup()` it should be picked up by HomeSpan.  What do you mean that you are "creating" HomeSpan?  Are you compiling from the IDF and creating your own tasks?

PR #997, opened 2025-01-25
Add useEthernet

mgeramb:
In my app does the autodection of ethernet now work.
The reason is, I have to start the network to get the KNX configuration. Depending on that, I have to create HomeSpan or not. But for this reason, Ethernet is already started and the event for starting is already sent, and HomeSpan does not get the start event.

### DIFF
--- a/src/HomeSpan.cpp
+++ b/src/HomeSpan.cpp
@@ -101,7 +101,7 @@ void Span::init(){
   
   networkEventQueue=xQueueCreate(10,sizeof(arduino_event_id_t));    // queue to transmit network events
   Network.onEvent([](arduino_event_id_t event){xQueueSend(homeSpan.networkEventQueue, &event, (TickType_t) 0);});
-  Network.onEvent([](arduino_event_id_t event){homeSpan.ethernetEnabled=true;},arduino_event_id_t::ARDUINO_EVENT_ETH_START);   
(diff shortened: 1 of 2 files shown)


mgeramb:
Sorry, for closing:  https://github.com/HomeSpan/HomeSpan/pull/994
This happens unplanned after an rebase. 
> Did you still need this function - I noticed the commit was pulled back. Also, can you provide some more details on exactly how you are using HomeSpan? In the dev branch the event handler is configured before the main Arduino setup() function is called, so as long as you are configuring your ethernet board inside setup() it should be picked up by HomeSpan. What do you mean that you are "creating" HomeSpan? Are you compiling from the IDF and creating your own tasks?

Yes, I still need this function. It is for this project https://github.com/OpenKNX/OAM-SmartHomeBridge.
This is an KNX bridge for different smart home systems. Currently it supports HomeKit and a Philips HUE Emulation.
The user can configure in the KNX-configuration tool ETS, which kind of protocol should be used. 
I figured out, that in my case it seems, that the creation of homespan happens to late, maybe because the initialization happens in an helper class https://github.com/OpenKNX/OFM-SmartHomeBridge/blob/v1dev/src/HomeKitBridge.cpp an the flag ethernet enabled is not set or maybe set to late.
This is the reason why I have added the useEthernet function which allows me to set the flag manually depending on the Hardware definition (KNX_IP_LAN).

PR #1021, opened 2025-02-24
Gracefully fail and abort on SPIFFS OTA

frankonski:
This will gracefully fail an OTA update of the SPIFFS, else the OTA process completes and fails on HomeSpan magic cookie validation.


frankonski:
I'll create an issue with steps to reproduce, as I cannot reproduce it at will...

frankonski:
Cancelling, something else is going on.

Issue #1023, opened 2025-02-26
OTA upload of SPIFFS fails after reflash


frankonski:
Doing an OTA upload of a filesystem image (SPIFFS) fails after a reflashing with valid filesystem image already uploaded via serial.

Error logged via the serial output:
```
*** Current Partition: app0
*** New Partition: app1
*** OTA Starting..10%..20%..30%..40%..50%..60%..70%..80%..90%..100%..Checking for HomeSpan Magic Cookie: �������
������������������������������������������������������������
������������������������������������������������������������
������������������������������������������������������������
������������������������������������������������������������
���������E�5*␄..*** OTA Error[4]: End Failed

[247933][E][ArduinoOTA.cpp:369] _runUpdate(): Update ERROR: Aborted
```
Steps to reproduce, using PlatformIO project tasks:

With `platformio.ini` set to upload via COM4, 
1. Erase Flash and Upload
2. Upload Filesystem Image
3. Connect to WiFi

At this point, I have a working setup that reads a json config file from SPIFFS
Edit `platformio.ini`, set to upload via espota (and set upload port / flags / auth)
4. Upload Filesystem Image OTA

This is where you get the error, and the board gets into an abort state.

PR on its way, to do the _Homespan Magic Cookie_ check only when the OTA in progress is for a new partition flashing.
Error is happening because there is no 2nd partition yet.

HomeSpan:
Update based on PR is available in dev.

frankonski:
gracias señior!

HomeSpan:
This has now been rolled out to production (HomeSpan version 2.1.2).

PR #1024, opened 2025-02-26
Check new partition magic cookie only when OTA is for new partition.

frankonski:
To address issue #1023, when uploading SPIFF while 2nd OTA partition is empty.

### DIFF
--- a/src/HomeSpan.cpp
+++ b/src/HomeSpan.cpp
@@ -2736,7 +2736,7 @@ void SpanOTA::progress(uint32_t progress, uint32_t total){
     LOG0("%d%%..",percent);
   }
 
-  if(safeLoad && progress==total){
+  if(safeLoad && progress==total && ArduinoOTA.getCommand() == U_FLASH){
     SpanPartition newSpanPartition;   
     esp_partition_read(esp_ota_get_next_update_partition(NULL), sizeof(esp_image_header_t) + sizeof(esp_image_segment_header_t) + sizeof(esp_app_desc_t), &newSpanPartition, sizeof(newSpanPartition));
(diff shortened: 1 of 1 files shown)


HomeSpan:
I understand what the pull request does but I don't understand what it is supposed to cure or what you were trying to do.  Were you uploading SPIFFs via OTA through the HomeSpan interface?  If so, that's neat - I did not think it would allow that without modification.  And of course since you are uploading SPIFFS, not an OTA partition, we can't check for magic cookie in the "next" partition, since the next one was not uploaded (SPIFFS were uploaded), and your pull request makes sense.

However, based on #1023 I got the impression that you were using USB to upload SPIFFS, and then separately, afterwards, you try to upload a sketch via OTA and it failed due to the magic cookie.  That doesn't make any sense to me - the magic cookie comes across IN the sketch you upload, directly into the partition into which the file is being uploaded.  The code then check that "next" partition and should find it.

Can you clarify - are you uploading SPIFF via OTA, or via USB?

frankonski:
Yes, uploading SPIFFS via OTA through what seems to be Arduino OTA which HomeSpan uses. I was happily surprised that it worked out of the box (except that small hiccup).

The `espota.py` / `espota.exe` tool supports the `—spiffs` option, and that is what’s used by PIO (and Arduino IDE??).

And yes, I initially upload the spiffs via USB as this steps seems to create the filesystem and format the spiffs partition properly before writing the two json files I place there. If I don’t do that and go straight to trying to an OTA upload of a spiffs after a full flash erase I get some kind of error like “no filesystem found”. I can get the exact error if you’d like.

Steps are:
- Full flash erase
- Upload HS partition via USB
- Upload spiffs via USB
- Try to upload spiffs via OTA. This is where is fails and what this PR is about.

Issue #1081, opened 2025-06-13
Support IPv6


dkerr64:
I had a PR submitted to my RAGTDO32 project that enabled IPv6 support.  It is as basic as calling `WiFi.enableIPv6()` and in general "it just works" except it turns out it doesn't.

I believe that the issue is that we can have multiple IPv6 addresses assigned by a network, and that results in multiple callbacks to HomeSpan, specifically [here](https://github.com/HomeSpan/HomeSpan/blob/0a998cdccc22b0c3847e0de4462ffd221053c6af/src/HomeSpan.cpp#L467).  And every time that is called, the `connected` global is incremented.  The problem is that if it is odd, then the connection is assumed to be good, but if even it is assumed to be disconnected... which causes a loop of WiFi connection attempts.

So my ratgdo32 logs look like this (stripped down to just the relevant messages)
```
I (5.947) ratgdo-homekit: Status: WiFi connecting
Garage Door CDFB80 is READY!

Trying to connect to KSec.  Waiting 5 sec...
I (5.958) ratgdo-homekit: Wifi begin for SSID: KSec
WiFi Connected!  IP Address = 192.168.21.89   (RSI=-43  BSSID=26:5A:4C:51:32:51)

Starting MDNS...

HostName:      Garage-Door-CDFB80.local:5556
Display Name:  Garage Door CDFB80
Model Name:    ratgdo-ESP32
Setup ID:      udm4

Starting HAP Server on port 5556...


DEVICE NOT YET PAIRED -- PLEASE PAIR WITH HOMEKIT APP

I (8.878) ratgdo-homekit: WiFi established, count: 1, IP: 192.168.21.89, Mask: 255.255.255.0, Gateway: 192.168.21.3, DNS: 192.168.21.3 Link_IPv6: ::
I (8.895) ratgdo-homekit: Found 0 IPv6 addresses:
WiFi Connected!  IP Address = 192.168.21.89   (RSI=-43  BSSID=26:5A:4C:51:32:51)
I (10.391) ratgdo-homekit: WiFi established, count: 1, IP: 192.168.21.89, Mask: 255.255.255.0, Gateway: 192.168.21.3, DNS: 192.168.21.3 Link_IPv6: fe80::da13:2aff:fecd:fb80
I (10.410) ratgdo-homekit: Found 1 IPv6 addresses:
I (10.415) ratgdo-homekit:   fe80::da13:2aff:fecd:fb80
[...]

PR #1082, opened 2025-06-17
Simplify Connection Logic

apearson:
Simplifying the connection logic from even and odd to just true and false with a connection counter. 

This should solve #1081 and allow HomeSpan to receive any number of IPv6 addresses without assuming the connection dropped. 

## Changes
- Switched from connected(int) to connected(bool)
- Created new connectionCount
- Using `ARDUINO_EVENT_WIFI_STA_CONNECTED` and `ARDUINO_EVENT_ETH_CONNECTED`  to detect when we connect 
- Updated HAP.cpp with new logic 

## Testing
I've imported and built a version of [homekit-ratgdo32](https://github.com/ratgdo/homekit-ratgdo32) with this branch and it looks like the disconnections have gone away. 

Screenshot 2025-06-16 at 10 20 07 PM

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,6 @@
+{
+  "files.associations": {
+    "stdexcept": "cpp",
+    "string": "cpp"
+  }
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,0 @@
-{
-  "files.associations": {
-    "stdexcept": "cpp",
-    "string": "cpp"
-  }
-}
--- a/src/HAP.cpp
+++ b/src/HAP.cpp
@@ -1125,7 +1125,7 @@ void HAPClient::getStatusURL(HAPClient *hapClient, void (*callBack)(const char *
     hapOut << "<tr><td>Compile Time:</td><td>" << homeSpan.compileTime << "</td></tr>\n";
 
   if(!homeSpan.ethernetEnabled){
-    hapOut << "<tr><td>WiFi Disconnects:</td><td>" << homeSpan.connected/2 << "</td></tr>\n";
+    hapOut << "<tr><td>WiFi Disconnects:</td><td>" << homeSpan.connected/2 << "</td></tr>\n"; // TODO: change this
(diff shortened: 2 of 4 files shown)


apearson:
> an you provide some color on why you added IP6 

I have an dual stack network (IPv4 + IPv6) and the only device that does not use IPv6 is my ratgdo32 (using [homekit-ratgdo32](https://github.com/ratgdo/homekit-ratgdo32) which depends on HomeSpan). Once this device is IPv6 I can move IPv4 to an optional thing that my devices can use if they want but I don't need to manage. I opened a PR with ratgdo32 that enables the ESP32 board to use IPv6 and it mostly just worked except HomeSpan having an issue where getting an even number combo of IPv4 and IPv6 addresses causes the library to believe it's in a disconnected state.

>  Why enable IP6 if the device is going to get an IP4 and IP6 address? 

IPv6 can be used by IPv6 clients and IPv4 can be used by IPv4 clients. You could also say this is future proofing ratgdo32 as everything is slowly moving to IPv6 (Matter).

>  Should IP4 connect be ignored by HomeSpan if the user enables IP6?

No, we should support both IPv4 clients and IPv6 clients simultaneously.  Some clients might be IPv6 (homepods, iphones) and some clients might be IPv4 only. 

> did not realize that the connection logic could be called twice (once for IP4 and once for IP6).

I think there's still a slight disconnect.  The `ARDUINO_EVENT_WIFI_STA_GOT_IP` mostly can be called once (maybe again if DHCP changes the IPv4 address) but `ARDUINO_EVENT_WIFI_STA_GOT_IP6` in most cases will be called multiple times (link local, global IPv6, ULA) and may result in this event being called an even or odd amount of times (and their eth counterparts).  Also we shouldn't ignore future events as IPv6 addresses might change (ISP prefix changes) and a downstream user might want to know when there is a new addresses.

Let me know if any of this doesn't make sense!

HomeSpan:
I'm trying to test out IPv6, but can't seem to get Safari to link to an IPv6 address to call the HomeSpan weblog once IPv6 is enabled.  How are you testing?

PR #1086, opened 2025-06-27
HTML compliance fix. Make sure html meta tag is in head tag.

frankonski:
While looking at initial access point connection issues I noticed the HTML `meta` tags were not in `head`. This PR improves html compliance.


HomeSpan:
Thanks for this.  Though I did not take the PR, I cleaned up the HTML code a bit and ensured the meta tag was always placed inside `<head>` as you correctly noted.   I also added a `<title>` tag (which is also required by HTML) and made sure the charset specifying UTF-8 was set in both the response header as well as in a meta tag (as appears to be the recommendation).

Update is available in the dev branch for testing.

frankonski:
Cool thanks. Feel free to close as done.

PR #1102, opened 2025-08-10
Use rmt_simple_encoder to directly generate rmt_symbols from Pixel::Colors

rubillos:
HomeSpan Version: 2.1.3
Arduino-ESP Ver.: 3.2.0
ESP-IDF Version:  5.4.1
ESP32 Chip:       ESP32-S3 Rev 2 dual-core 16MB Flash
ESP32 Board:      um_feathers3

When driving a set of 4 LEDs using the Pixel class, the output waveform had gaps between the pulses for the individual pixels that was long enough to trigger a reset in the stream, so only the first LED was being updated. I looked at the waveform being generated on an oscilloscope and the gap between the pixels is almost as wide as an entire pixel's bits. Seems like calling rmt_transmit multiple times in a row introduces delays.

I initially modified the code in Pixel::Set to create a small buffer for the data for all of the pixels and then sent them in a single rat_transmit call, which fixed the problem. In my case, with just 4 LEDs, the buffer wasn't a big deal but I was interested to know if a rmt_simple_encoder could be used to encode Color data directly into rmt_symbols. The new code works and does not use any buffer memory to send to any length of LED strip.

### DIFF
--- a/src/src/extras/Pixel.cpp
+++ b/src/src/extras/Pixel.cpp
@@ -38,18 +38,18 @@
 IRAM_ATTR size_t Pixel::pixelEncodeCallback(const void *data, size_t data_size,
                      size_t symbols_written, size_t symbols_free,
                      rmt_symbol_word_t *symbols, bool *done, void *arg) {
-  Pixel::rmt_pixel_encoder_config_t* encoder_config = (Pixel::rmt_pixel_encoder_config_t *)arg;
+  rmt_pixel_encoder_config_t* encoder_config = (rmt_pixel_encoder_config_t *)arg;
   Pixel* pixel = encoder_config->pixel;
-  Pixel::Color* colors = (Pixel::Color *)data;
+  Color* colors = (Color *)data;
   int bytesPerPixel = pixel->bytesPerPixel;
   int symbolsPerPixel = 8 * bytesPerPixel;
-  int totalPixelCount = data_size / sizeof(Pixel::Color);
+  int totalPixelCount = data_size / sizeof(Color);
   int pixelsWritten = symbols_written / symbolsPerPixel;
   bool multiColor = encoder_config->multiColor;
   size_t symbolsGenerated = 0;
 
   while (pixelsWritten < totalPixelCount && symbols_free >= symbolsPerPixel) {
-    Pixel::Color* srcColor = colors + (multiColor ? pixelsWritten : 0);
+    Color* srcColor = colors + (multiColor ? pixelsWritten : 0);
 
     for (auto i=0; i<bytesPerPixel; i++) {
       uint8_t byte = srcColor->col[pixel->map[i]];


HomeSpan:
Are you using RGB or RGBW LEDs - have you run the PixelTester Example to double-check?

rubillos:
I'm using RGB LEDs. I did not run the PixelTester. When I initially just changed the code from doing 1 rmt_transmit per pixel to doing a single rmt_transmit for all 4 pixels, all 4 LEDs showed the proper colors, rather than the 1st LED showing the color for the last LED, which is what I initially saw.

HomeSpan:
Yes, that is as intended.  You should send a single transmit for all 4 pins at the same time.  If they are all the same color, you can specify that color.  If you want each pixel to be a different color/brightness, you can send an array of colors and the Pixel library will handle accordingly.  Every transmit is considered a call to set all Pixels in a strand.  This minimizes CPU time by maximizing the use of the ESP32 RMT peripheral.

rubillos:
My 4 LEDs are hooked up in a string in series, connected to a single data pin. With the existing Pixel::set method, each pixel's data is re-mapped into a local buffer and sent via an rmt_transmit, one after the other. In my case this causes 4 rmt_transmit calls to happen for each call to Pixel::set. Between each call there is a delay introduced into the signal stream on the output pin (I measured approximately 30us), which is long enough to be considered a reset by the LEDs. This causes the first LED to think the bitstream is starting over after each pixel is sent.

rubillos:
What I see on the output pin when I send 4 pixels is that there are 30us gaps _between_ the data for each pixel. So for a single Pixel::set call with and array of 4 Colors and a pixelCount of 4, the output on the pin is:

24 pulses - 30us gap - 24 pulses - 30us gap - 24 pulses - 30us gap - 24 pulses

The gaps should not be there, it should be a continuous stream of 96 pulses.

HomeSpan:
Sorry, I cannot reproduce the issue you are having.  I'm testing an 8-pixel strand using an ESP32-S3 running HomeSpan 2.1.3 under Arduino-ESP32 3.2.0 as per your specs above.  I can set all 8 colors in one call as well as set them all to different colors using a Color array (with a single call passing the array).

Can you post a barebones copy of your sketch that reproduces the issue.

rubillos:
Let me re-investigate tomorrow. I'll roll back my code and get you a sketch and a snapshot of what I see on the oscilloscope.

HomeSpan:
I don't have a scope so can't measure the signal timing.  However, the sketch runs fine on my Pixel strand.  It lights up 4 Pixels, each with a different color as per the sketch (red, green, blue, yellow). 

Can you try running on a different pin, or maybe even a different chip.  I'm not sure why that would matter but I'm at a loss to explain what you are seeing.

Note your pull request is not backwards compatible with prior Arduino-ESP32 versions, which I try to maintain whenever possible.  Espressif CONSTANTLY tinkers with their API and creates a lot of breaking changes.  It's a huge pain and I try NOT to pass them on to users.  When Espressif released version 5 of the IDF they re-did the entire RMT API, which meant that when they then released Arduino-ESP32 version 3, which uses IDF 5, I had to completely re-write the HomeSpan Pixel and RFControl libraries.  At that time, the Arduino-ESP2 library pointed to IDF 5.1, which does *not* include the `rmt_simple_encoder`.  This was not added until a later IDF version and was therefore not available in the first few versions of the Arduino-ESP32 library.  Hence, I needed to write my own driver.  This was frustrating since the `rmt_simple_encoder` was already available by that time in the new IDF version, and it seemed like a obvious choice, but Espressif had not yet updated the Arduino-ESP32 library to point to the latest IDF (would take many more months for that to occur) so I could not use it.

Hence, if there is an issue with the current Pixel library, I'll need to address it in the current framework without using `rmt_simple_encoder` else HomeSpan will not compile on earlier versions of the Arduino-ESP32 library.  And I know there are lots of users who use older versions and for various reasons can't update (likely since they don't want to deal with all the breaking changes Espressif introduces each time).

HomeSpan:
One final thing to double-check.  Are you using the production current production version of HomeSpan 2.1.3 (the official release)?

HomeSpan:
I have finally been able to reproduce the issue by judicious use of print statements measuring microseconds peppered through the Pixel code.   The problem is a fixed overhead internal to the rmt_transmit when using the byte encoder.  I had assumed it was simple pass-through but alas there is very significant overhead each call.

My goal in creating the Pixel library was to minimize RAM usage --- there should be no reason to store an array of 1000 colors more than once, anywhere.  If you send 3000 bytes to rmt_transmit, it creates a new 3000 byte array that it then sends to the RMT peripheral using the ping-pong method already in the Pixel library.   So instead I would call rmt_transmit 1000 times with 3-bytes each.  This saves a a bunch of memory, especially if you just want to send the same color for each pixel.  But the overhead inside rmt_transmit appears to be very significant.

There are a few possibilities.  One is to use the new encoder (as you suggested).  The second is to expand the queue used in the rmt_transmit (for the byte encoder) to a much larger value.  This uses more memory but avoids the overhead of calling rmt_transmit more than once. 

I'll give some thought to the different solutions and implement something in the next few weeks.

The one question I cannot answer is why the current method seems to work for so many Pixel strands - I think it must mean that the 30 us overhead is not long enough to re-trigger a reset on certain strands, but is long enough to trigger a reset on others.

Thanks for helping identify this latent issue (present since I updated the library in HomeSpan 2.X)!

rubillos:
Glad you were able to reproduce it! I did some digging while I was researching and found this article:
[NeoPixels Revealed: How to (not need to) generate precisely timed signals](https://wp.josh.com/2014/05/13/ws2812-neopixels-are-not-so-finicky-once-you-get-to-know-them/)
It mentions there being quite a bit of variation in how long a gap is recognized as a reset by different brands of LEDs. Perhaps my LEDs are just at the lower end? (I did need to adjust the timing parameters for my LED strips, setTiming(0.35, 0.8, 0.7, 0.6, 80.0) is what worked for mine, the defaults caused flickering.

I initially tried modifying the code to do a single rmt_tx_wait_all_done, then doing all of the rmt_transmits, and then another rmt_tx_wait_done. Since queue_nonblocking is not set in rmt_transmit_config_t, I expected the rmt_transmits to queue up, to the limit of 4 that's set, then progress as the transmits completed, which seemed like it should solve the issue, but that didn't help the problem. I might take another look at that possible solution, maybe I made an error in the attempt. Perhaps looking at the result code returned from rmt_transmit and resending if it fails.

HomeSpan:
I've given a lot more thought to this and tried a few more things using the byte_encoder but no matter what there is just too much overhead in rmt_transmit to call more than once per pixel update.   Though it will mean that user will need to update to at least version 3.1.0 of the Arduino-ESP32 library, I think using the simple_encoder is going to be best in the long-term.  I'm going to accept your PR into a new branch to do some testing (I have some long RGBWC pixel strands) and if all goes well will add to dev branch for inclusion in the next release.  Will let you know if anything surfaces in testing.

rubillos:
Great! That should also help with long strips as the overall transmission will be something like twice as fast without the gaps.

HomeSpan:
I've thoroughly tested your PR and found a bug - *NOT* in your code, but in the IDF itself.  There is a problem in the simple_encoder logic that does not properly account for all space needed to tack on the EOF when a transmission is done.  I stumbled onto this while testing an RGBW strand.  In your tests on the S3 using an RGB, there are 24 bits per Pixel, and 48 symbols in the RMT hardware memory (per channel), so things fit nicely without simple_encoder needing to use its internal cache.

But for an RGBW with 32 bits per pixel, simple_encoder uses an internal cache.  This worked when using an even number of pixels, but the last byte got dropped when using an odd number of pixels.  This took a while to identify and debug.  The solution I found was to change the callback logic so that you set *done to true AFTER your last transmission, instead of at the same time.  Both are allowed according to the IDF docs, but sending in a subsequent call ensures all bytes are transmitted during the prior call.

Apparently, I was not the only one who found this issue - it was recently reported to Espressif as well (who confirmed the bug).

In any case, the new logic works fine and is ready for testing in the *newPixel* branch.  I also wound up making a number of changes to the PR code to match the style of the rest of HomeSpan, and I optimized the callback logic (to a certain extent) to minimize computations since we are using an interrupt to transmit pixels.

Please give it a try and let me know if you find any issues.

PR #1106, opened 2025-08-19
Add a flag to SpanPoint to enable overwriting the last value in the queue.

Andrew-Hanlon:
While perhaps valid in certain crucial tracking/logging use cases, I believe the most common use case for SpanPoints are sensors where only the latest value is important. This pull request adds a flag that lets the SpanPoint overwrite the last value in the queue instead of failing. Perhaps this could be implemented in a different way, say a 'QueueBehaviour' enum, but this is a simple solution to the most common use case (queue length of 1, only needing the latest value).

### DIFF
--- a/src/HomeSpan.cpp
+++ b/src/HomeSpan.cpp
@@ -2923,7 +2923,7 @@ boolean SpanOTA::auth;
 //        SpanPoint          //
 ///////////////////////////////
 
-SpanPoint::SpanPoint(const char *macAddress, int sendSize, int receiveSize, int queueDepth, boolean useAPaddress){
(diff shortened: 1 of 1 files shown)


HomeSpan:
Thanks.  Seems like a good idea - I'll review and promote to next release.

HomeSpan:
FYI - the very next version of HomeSpan (2.1.4) will be released shortly.  It provides a critical fix of the Pixel library.  Prior to needing to work on this fix, I had been working on updating SpanPoint to be able to use broadcast addresses to obtain the MACs of each device and create a secure network without needing to know the MACs ahead of time.  Updating SpanPoint will be the focus of version 2.1.5, so I'll add this PR to that version.

PR #1144, opened 2026-01-28
Improve Content-Length parsing & error handling, Reimplement Nonce::inc() for security

Xorlent:
HAP.cpp:
1. Rewrite Nonce::inc() to use all bytes (4-11) to prevent nonce reuse
2. Improved Content-Length validation and error handling
3. Fix potential buffer overflow when testing decrypted content length

Network.cpp:
1. During serial Wi-Fi configuration, verify broadcasted SSID string length to guard against overflow and prompt user to select a different SSID if necessary

HomeSpan.cpp:
1. Even if ETH interface is initialized, we need access to an entropy source, so we set Wi-Fi station mode during HomeSpan initialization.


Xorlent:
Hi Gregg, please let me know if you would prefer I submit this merge request against the dev branch.

Thanks!

PR #1147, opened 2026-02-07
Implement programmatic resource monitor access

Xorlent:
Add Utils::getSystemStatus() function for programmatic resource monitoring (see [Discussion 848](https://github.com/HomeSpan/HomeSpan/discussions/848))

- Created ResourceMonitor struct to return system metrics
- Includes heap memory (internal and PSRAM if available): allocated, free, largest block, low water mark)
- Includes active HAP socket connections (e.g. CLI 's' command)
- Includes stack high water marks for poll and loop tasks (e.g. CLI 'm' command)
- Required adding getActiveConnections() and getLoopTaskHandle() getters to Span class
- Returns -1 for pollTaskStack when autoPoll() is not enabled
- Example .ino in examples/Other Examples/SystemStatus/


HomeSpan:
@Xorlent, I really appreciate your enthusiasm in adding these new features but HomeSpan is presently used by thousands of people and at this stage stability is of paramount importance, as well as the ability for me to be able to continue supporting the system.  There is a rhyme and reason for all the functions, how they fit together, why they exist, etc.  

For instance, based on the specific request to see socket connections to determine if a reboot may be necessary, it might appear that adding a function to return that info would be the solution (along with perhaps other data as you've added).  But the way one would most likely use this is to detect the extended absence of a secure HomeKit connection --- say for 30 or 60 minutes.  Which means you would have to poll the connections and keep track of how long the device had no connections before rebooting.  Also, we need to distinguish connections from HomeKit versus WebLog, since the WebLog sockets can stay active for a while after an HTTP request. 

The solution here is not to add new functions, but to extend the current design of existing functions.  HomeSpan already has a callback to monitor state changes in the system `setStatusCallback()`.  States progress from no-wifi credentials, to wifi credentials found but can't connect, to wifi connected but device not yet paired, to wifi connected and device paired.  To address the user's goal of rebooting when connections are lost in a way that is consistent with the current design of HomeSpan, I was going to add a new state: wifi connected, device paired, AND device has at least one HomeKit connection. 

By extending this existing function, users would be able to monitor state changes (as they do now) and save the time when state changes occur (might even add that as a second argument to the callback).  Based on the last time of an unwanted state change, the user can then reboot if desired.  As a bonus, since state changes are already coded to the Status LED, users would also be able to visually see if there were no HomeKit connections by simply looking at their device.

A core design philosophy is also to try to minimize functions that are basically duplicative of existing IDF functions --- I don't always do a good job of that, but software bloat is something I try to fight whenever possible.   If users need a function to return heap memory, high-water marks, etc., they should use the existing IDF functions.  Users who need such info typically have something advanced they are trying to do, and they may need to distinguish the different types of memory available (eg, IRAM vs DRAM) and the IDF already has all options built-in, so it's best to use that function directly instead of adding similar functions to HomeSpan.

With regard to the original sockets question, I also intend on adding a duplicate of the 's' output to the WebLog so it is possible to see exactly what is connected in real time, which my be helpful in debugging network issues.

If you look through the history of PRs, you may notice that I don't adopt most of them --- it's just not practical for me to debug and support the variety of coding styles and implementations.   It's one of the reasons I created my own LED, Pixel, RF_Control, etc. drivers.  I originally just wanted to use existing libraries from others, but these libraries have a life of their own, change frequently, and have a host of other issues that the owners of those libraries are always asked to resolve.  I wanted to avoid disappointing users who have issues with how a Pixel LED works by telling them I can't do anything since it's not my code and they would need to contact the owners of the other library.

Contributions to HomeSpan by users are _critical_ to the success of the system.  _However_, the best way to contribute is by opening issues (for problems) or discussions (for ideas/features) that we can discuss in detail to find the optimal solution before any code is written.  That may lead to a coding contribution through a PR, but often it is something I can integrate more seamlessly in the next release.  I have found this to be the most productive, and supportable, path since first releasing HomeSpan a little over 5 years ago.

On a related note, thanks for catching the issue with my Nonce structure!  I recall when I first started testing the HAP logic, I created the Nonce structure the way I did to test whether the Nonce needed to be big-endian or little-endian.  The public HAP documentation (which Apple has long since deleted) is abysmal.  However, after I got it working I apparently never went back and coded it properly.  I've done so now and you can see the changes in the latest dev branch (I use a padded structure with a uint64_t for the non-zero portion of the nonce aligned so all 96-bits can be output as a uint8_t *).

I'll also shortly review the other changes you've suggested in that same PR.

Xorlent:
Hi @HomeSpan,

Thank you so much for taking the time to write such a detailed response, Gregg.  I really appreciate you diving into the PR and sharing your perspective and I hope this didn’t take you away from more important work.

My goal for creating a PR was actually to provide demonstration code for @phamvanan88 and others that may want/need temporary debug features.  Creating a PR allowed me to show the changes side-by-side against the current release so it would be easy to follow if the desire was there.  Once a user has identified and resolved their specific issue, they can return to mainline.  I suppose I could have linked to a fork with the edits in the discussion, but if I didn’t keep the fork synchronized the edits will start to diverge from what’s in the current release.

I completely understand your concerns about PR hygiene and noise, though, and it’s a very valid point, especially given how mature and stable the library is at this point.  In the future I will find other ways to illustrate example code that may help others start to more deeply interact with the code and libraries they use.

No thanks needed on the Nonce item – I’m glad I was able to contribute in some (albeit minor) way!

Xorlent:
Closing PR.

PR #1163, opened 2026-07-01
[feat] Add Apple HomeKit Adaptive Lighting support

nbelles:
## Summary

Adds Apple HomeKit **Adaptive Lighting** support to HomeSpan. Once this PR is merged, any `Service::LightBulb` that exposes `Brightness` and `ColorTemperature` becomes eligible for the Adaptive Lighting toggle in the Home App — the Home Hub (HomePod / Apple TV) uploads a 24-hour color-temperature schedule, and the ESP32 automatically shifts its color temperature over the day with no further intervention.

The Adaptive Lighting protocol is undocumented by Apple; this implementation follows the reverse-engineered spec used by [HAP-NodeJS](https://github.com/homebridge/HAP-NodeJS) and has been verified against traffic from real HomeKit controllers on an ESP32-C6 test device.

## User-facing API

Only one line of user code is needed to add Adaptive Lighting to an existing LightBulb Service:

```cpp
struct MyLight : Service::LightBulb {
  SpanCharacteristic *power       = new Characteristic::On(0);
  SpanCharacteristic *brightness  = new Characteristic::Brightness(100);
  SpanCharacteristic *colorTemp   = new Characteristic::ColorTemperature(250);

  AdaptiveLightingController adaptiveLighting{brightness, colorTemp};   // <-- new

  boolean update() override { adaptiveLighting.handleUpdate(); /* ... */ return true; }
  void    loop()   override { adaptiveLighting.poll(); }
};
```

The controller instantiates the three (undocumented) HAP characteristics required by the protocol automatically, so sketches don't have to touch them directly.

Additional public methods: `disable()`, `isActive()`, `printSchedule()`, `setColorTemperatureCallback(cb)`, `setStateCallback(cb)`.

## What's in the PR

**Library core**
- `src/Characteristics.h` — 3 new HAP characteristics:
  - `SupportedCharacteristicValueTransitionConfiguration` (0x144)
  - `CharacteristicValueTransitionControl` (0x143)
  - `CharacteristicValueActiveTransitionCount` (0x24B)
- `src/Span.h` — matching `CREATE_CHAR` entries in `namespace Characteristic {}` and matching `OPT(...)` entries in `Service::LightBulb`
- `src/AdaptiveLighting.h` — new header-only `AdaptiveLightingController` class (~600 lines including comments) implementing the full protocol: `UPDATE` / `READ_CURRENT` control-point handling, curve interpolation on the update-interval, brightness-adjusted mired output, EV notifications on the notify-interval, automatic self-disable on manual `ColorTemperature` writes
- `src/HomeSpan.h` — `#include "AdaptiveLighting.h"` after `Span.h`

**Example**
- `examples/23-AdaptiveLighting/23-AdaptiveLighting.ino` — minimal LightBulb + AL controller with two `SpanUserCommand` hooks (`@p` prints the currently-loaded schedule, `@x` cancels it) and inline commentary about NTP-based wall-clock time

**Docs**
- `docs/AdaptiveLighting.md` — new full documentation page (protocol overview, requirements, class reference, sample sketch, `printSchedule()` output description, wall-clock notes, caveats)
- `docs/Reference.md` — `AdaptiveLightingController` class entry
- `docs/Tutorials.md` — Example 23 entry
- `docs/ServiceList.md` — regenerated via `tools/makeServiceList` (includes an incidental typo fix `indictes` → `indicates` that fell out of the regeneration)
- `README.md` — link to the new doc page from the documentation list

## Design notes

- **Header-only.** `AdaptiveLighting.h` uses `inline` methods so there's no additional compilation unit. Trivial to move to a companion `.cpp` if preferred.
- **No behavioral changes to existing Services or Characteristics.** The three new characteristics are additive `OPT` entries on `Service::LightBulb`; every existing sketch continues to work unchanged.
- **Manual override.** If the user changes `ColorTemperature` from the Home App, the controller detects the write in `handleUpdate()` and cancels the schedule (matching Apple's behavior). An internal guard prevents the controller's own writes from being interpreted as manual overrides.
- **Time handling.** The controller anchors the schedule at the moment it is received and advances it using `millis()`. This is intentionally dependency-free; the doc page explains how to align with wall-clock time using `configTime()` + NTP if desired.
- **TLV8 same-tag records.** Zero-length separator records are inserted between consecutive same-tag records in both the outgoing supported-config and the `READ_CURRENT` response, matching the pattern used by the existing `DisplayOrder` support in Example 22.
- **No new dependencies.** Uses only the existing HomeSpan `TLV8`, `SpanCharacteristic`, and `LOG0/1/2` machinery, plus `<vector>`, `<functional>`, `<cmath>`, `<cstring>` from the standard library.

## Testing

- Compiled clean (no new warnings) for `esp32:esp32:XIAO_ESP32C6:PartitionScheme=huge_app`, arduino-esp32 core 3.3.8.
- Regression compile of `examples/06-DimmableLED` against the modified library — still clean.
- End-to-end verified against a real HomeKit setup (HomePod hub + iPhone + XIAO ESP32-C6 driving a WWA NeoPixel bulb): schedule uploads from the Home App are accepted, mired values track the curve over the update-interval, manual `ColorTemperature` writes correctly disable the schedule, `printSchedule()` output matches the wall-clock times displayed in the Home App.

## Not included

- **No version bump.** `library.properties` and `src/version.h` are unchanged. Happy to bump to 2.2.0 (or wherever you prefer) as part of this PR — just let me know.
- **No `Latest Update` entry in README.md.** Same reason — that's easier to add if/when this is being cut into a release.
- **No persistence of the active schedule across reboots.** The Home Hub re-uploads the schedule the next time it contacts the device, which matches how Homebridge behaves.

## Checklist

- [x] Compiles clean on ESP32-C6 with arduino-esp32 3.3.8
- [x] No API-breaking changes to existing Services / Characteristics
- [x] Example sketch added
- [x] Documentation added and cross-linked
- [x] `docs/ServiceList.md` regenerated with `tools/makeServiceList`
- [x] Tested end-to-end against a real Home Hub


HomeSpan:
Thanks very much for putting this together!  Would be great to add adaptive lighting to HomeSpan, especially since Apple has not provided any support for the required characteristics.

As for the PR itself, I think nearly all of the logic should be moved to a sketch (aka the application layer).  As a library, HomeSpan classes are designed to be primitives with everything needed to support the logic within a sketch that will act on Services, Characteristics, etc.  The AdaptiveLighting class should be embedded logic in the sketch as well, since it operates on a Service and is not by itself a Service.  What would belong in HomeSpan is of course the definitions of the Characteristics needed by HomeKit to support adaptive lighting for the LightBulb Service (as you correctly added them in the PR).

That said, before adding to the library, they need to be fully tested and vetted, which is one of the reasons why HomeSpan allows users to add custom Service and Characteristic definitions directly to their sketches.

My recommendation would be to first start with a standalone sketch that defines the needed Characteristics using the CUSTOM_CHAR macro, implements your AdaptiveLighting class, and then uses it within the LightBulb Service.  If you post that in your GitHub repository I (and any other users) would be able to test fully, clone as needed to make edits, etc.  Once all the parameters of the new Characteristics have been vetted I can then add them in permanently to HomeSpan and update the Characteristics documentation accordingly.

The sketch can then be converted into a self-contained HomeSpan example, and a new page can be added to the HomeSpan documentation describing the details of the Characteristics, the tag format, etc.  To the extent that others want to duplicate they can simply copy the example.   And since your AdaptiveLighting class is in the sketch, they will get the benefit of using it if they would like -- documentation for that class should be in the sketch itself.  But they could also modify it, change the logic, or use something entirely different in their own sketch, as long as they adhere to the details of the new Characteristics as described in the what would be the new docs.  This is the benefit of separating the core library functionality from the application layer.  Or at least that is how I hoped to design things.

It's also exactly how I added the Television Characteristics --- a user reverse engineered the Characteristics structure from which sample sketches were developed (for testing and then posting) and finally a documentation page.  However, the logic for how to use the Characteristics stayed in the sketch - only the new Characteristics themselves moved into HomeSpan.  This ensures the library stays manageable and supportable while providing maximum flexibility for anyone to customize functionality from within their own sketches.

nbelles:
To be honest, I probably won't spend time doing this anytime soon. Someone else is more than welcome to take what I did and run with it but it works for exactly what I need right now. I have a lot of other stuff going on that's higher priority for me and don't have a lot of free time to spend on getting this integrated directly with HomeSpan. I mainly was posting here for posterity so that others have something to start from if they also want to get Adaptive Lighting working without having to reverse engineer something. 

Thanks for all the work on HomeSpan to make this easy to get working!

HomeSpan:
No worries.  Really appreciate you building out the foundation for the adaptive lighting functionality.  I'll close the PR for now but will use as reference for future expansion.

PR #1165, opened 2026-08-02
Add ability to use a CSS file for the web log instead of inline CSS.

frankonski:
**Some context**

Using `homeSpan.getWebLog()` to get weblogs from a different URI, configured with a separate webserver listening on a different port, HomeSpan will generate the full HTML page. No issues here.

I now need to send the following HTTP header to protect against XSS:
```c
  httpd_resp_set_hdr(req,
    "Content-Security-Policy",
    "default-src 'self'; "
    "script-src 'self'; "
    "object-src 'none'; "
    "base-uri 'none'; "
    "frame-ancestors 'none'");
```

This now causes Chrome to abort the connection with this error:
```
Applying inline style violates the following Content Security Policy directive 'default-src 'self''. 
Either the 'unsafe-inline' keyword, a hash ('sha256-9uMUYPNHTeOWAevYKamF1ilLlQzsK3ALyCpUeBf/i5c='), 
or a nonce ('nonce-...') is required to enable inline execution. Note also that 'style-src' was not explicitly set, 
so 'default-src' is used as a fallback. The action has been blocked.
```

The error is caused by the inline `<style>...</style>` in the HTML generated by HomeSpan.

To address this issue, and make HomeSpan more secure, the `setWeblogCSS()` method has been enhanced with an optional parameter, `isURI`, defaulting to `false`. If set to `true`, the string passed to `setWeblogCSS()` is expected to be a URI, full or relative, to a .css file.

**Example:**
Prior to the change, I used this:
```c
homeSpan.setWebLogCSS("</style><link rel=\"stylesheet\" href=\"/static/style.css\"><style>");`
```
which generated this:
```html
<style>body {background-color:lightblue;} th, td {padding-right: 10px; padding-left: 10px; border:1px solid black;}
</style><link rel="stylesheet" href="/static/style.css"><style>
</style></head>
```

Now with the change, we can simply call this:
```c
homeSpan.setWebLogCSS("/static/style.css", true);
```

And HomeSpan will generate this:
```html
<link rel="stylesheet" href="/static/style.css" type="text/css" />
```

The security warning no longer appears.

### DIFF
--- a/src/HAP.cpp
+++ b/src/HAP.cpp
@@ -1106,8 +1106,12 @@ void HAPClient::getStatusURL(HAPClient *hapClient, void (*callBack)(const char *
 
   if(homeSpan.webLog.faviconURL)
     hapOut << "<link rel=\"icon\" href=\"" << homeSpan.webLog.faviconURL << "\" type=\"image/png\" />\n";
-    
-  hapOut << "<style>body {background-color:lightblue;} th, td {padding-right: 10px; padding-left: 10px; border:1px solid black;}" << homeSpan.webLog.css.c_str() << "</style></head>\n";
+  
+  if (homeSpan.webLog.cssURI)
+    hapOut << "<link rel=\"stylesheet\" href=\"" << homeSpan.webLog.css.c_str() << "\" type=\"text/css\" />\n";
+  else
+    hapOut << "<style>body {background-color:lightblue;} th, td {padding-right: 10px; padding-left: 10px; border:1px solid black;}" << homeSpan.webLog.css.c_str() << "</style></head>\n";
+  
   hapOut << "<body class=\"body bod1\"><h2>" << homeSpan.displayName << "</h2>\n";
   
   hapOut << "<table class=\"infoTable tab1\">\n";
--- a/src/HomeSpan.h
+++ b/src/HomeSpan.h
@@ -224,6 +224,7 @@ struct SpanWebLog{                            // optional web status/log data
   char *faviconURL=NULL;                      // optional URL for favicon PNG image
   uint32_t waitTime=120000;                   // number of milliseconds to wait for initial connection to time server
   String css="";                              // optional user-defined style sheet for web log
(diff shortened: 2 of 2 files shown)


HomeSpan:
A few questions/clarifications:

1) By itself, this does not make HomeSpan more secure.  That comes from running your own server for the pages with added security.  However, with that added security, Chrome won't process the style sheet if it's inline, but will if it is a link to a separate file.  This only works because you are serving your own pages and can store them (I assume in SPIFFS)?

2) Do you still get the warning message when you use
`homeSpan.setWebLogCSS("</style><link rel=\"stylesheet\" href=\"/static/style.css\"><style>");`?  If so I assume that's because you need to not only specify the style sheet as a static reference, but also get rid of the default inline style settings HomeSpan invoke sprior to the custom CSS?

3) Are there any other possible HTML structures one might want to use under different circumstances instead of `</style><link rel=\"stylesheet\" href=\"/static/style.css\"><style>"` where the only parameter you can specify is the href?  In other words, would it be more general to have a command that simply disables the HomeSpan `<style>` HTML output and you can then add whatever HTML code you need in your server to generate the style in a static file, or any other HTML code to perform any other functions?

frankonski:
Hi,

1. Correct across the board. I have a mix of static html, css, js files stored in SPIFFS, rendered by the browser as-is. But they contain DOM events to fetch & load dynamic data in JSON. I can deal with all these files on my own. But the fact that I have a URI handle (`/logs`) that uses the weblog callback means that this page now has issues when I define the CSP header. I'd prefer to keep everything as secure as possible.
2. Yes & Correct. This is how I had been applying the same style to all my pages, including weblogs. The problem is the `<style>` tag. I had started down the route to do a "string replace", but I thought it would be cleaner if I could get this PR approved.
3. Some kind of `homeSpan.disableCSS()` **and** `homeSpan.addHTMLHeader()`? Maybe. But I do not want to over-complexify your code meant to simply display logs. 

I'm open to any suggestion to get rid of `<style>` and allow me to add a `<link rel="stylesheet" href="/static/style.css" />` inside the `<head>..</head>` section.

